Issue #37, opened 2021-09-15
Are smoothing matrices correct?


brendanjmeade:
Jack- I've done the initial construction of sparse smoothing matrices.  They are the right size and seem to have sparse terms of the main diagonal.  However, I'm not certain that there any non-zero values off of the main diagonal, so I think I may have messed up somewhere.  Let me know if you might have time to take a look!

jploveless:
Any ideas why pasting this into a notebook cell:

```
meshes[0].share = celeri.get_shared_sides(meshes[0].verts)
print(meshes[0].share[0:10,0:3])
```

gives the correct result (with -1 in matrix entries where there is no shared side) and pasting:

```
celeri.get_all_mesh_smoothing_matrices(meshes) 
print(meshes[0].share[0:10, 0:3])
```

into the next cell gives the wrong result (non-empty indices are correct, but there are seemingly random integers where there should be -1)?

PR #38, opened 2021-09-16
Fixes to triangular smoothing matrix

jploveless:
This seems to fix the triangular smoothing matrix from #37. The new results, plotted in the notebook, are similar to those from the Matlab equivalent (though smaller in magnitude since distance scaling in celeri uses meters, not km). There were some funny indexing things that stemmed from changing meshes.share within the distance calculation, but I just do the distance calculation differently so that it avoid changing anything about meshes.share.

### DIFF
--- a/celeri.ipynb
+++ b/celeri.ipynb
@@ -11,19 +11,12 @@
    ],
    "outputs": [],
    "metadata": {
-    "execution": {
-     "iopub.execute_input": "2021-08-22T18:29:50.539195Z",
-     "iopub.status.busy": "2021-08-22T18:29:50.538938Z",
-     "iopub.status.idle": "2021-08-22T18:29:50.550334Z",
-     "shell.execute_reply": "2021-08-22T18:29:50.549700Z",
-     "shell.execute_reply.started": "2021-08-22T18:29:50.539169Z"
-    },
     "tags": []
    }
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 62,
    "source": [
     "import addict\n",
     "import numpy as np\n",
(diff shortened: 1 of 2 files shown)


brendanjmeade:
Thanks, Jack!  This is working for me and now merged.  I just submitted another commit (https://github.com/brendanjmeade/celeri/commit/91dcecc7ae8cd4bc439b18024447ac01d8f2c636) that takes what you built and makes smoothing_matrix a sparse matrix.  I also used directing indexing with "k" and "m" to avoid the flattening and ravel approach, which I don't think is supported by sparse matrices.  I get what looks like the same answer!
Figure 12

jploveless:
That's interesting: direct indexing would not work in Matlab because it'd try to fill the full 3-by-3 block identified by the `k` and `m` arrays with the scalar value of `off_diagonal_terms[i, j]`. I'm not sure how Python knows to put that scalar along the main diagonal, but it definitely simplifies things vs. the ravel/flat approach. 

Sorry, I had made this a full matrix because I always had to call `full` in Matlab back when sparse matrix support was kinda minimal. If there are good tools in Python for inverting full/sparse combos, it'll be great to have the memory savings by keeping smoothing sparse.

brendanjmeade:
I'm not sure that there are good tools for inverting sparse matrices but I figured it was easier to just store them this way.  We can promote them to full numpy matrices with smooth_matrix.toarray().  With the SVD approach on the horizon I think the end game will to just calculate the triangular partials and smoothing matrix for one mesh at a time, do the weighted eigenvalue approach, and then discard the full triangular partials and smoothing matrix because we won't need them anymore.  Not having to store all of these all the time will save a huge amount of memory!

PR #41, opened 2021-09-20
Add new mesh parameters to fix CI failure.

tbenthompson:


brendanjmeade:
Thanks for this.  I really failed in debugging this because I didn't think.  I should have dug into the test and listened to what it was telling me.  Lesson learned.

PR #43, opened 2021-09-27
Fixing the cutde vs okada test.

tbenthompson:
I left some comments in the code that might help understand where the errors were.

### DIFF
--- a/celeri.py
+++ b/celeri.py
@@ -2317,14 +2317,15 @@ def test_okada_equals_cutde():
     # for every observation point.
     u_cutde = np.sum(disp_mat.dot(slip[0]), axis=2)
 
-    field_cutde = u_cutde[:,0].reshape((nx, ny))
-    field_okada = u_x_okada.reshape((nx, ny))
-    plt.subplot(1,2,1)
-    plt.contourf(x_obs_mat, y_obs_mat, field_cutde)
-    plt.colorbar()
-    plt.subplot(1,2,2)
-    plt.contourf(x_obs_mat, y_obs_mat, field_okada)
-    plt.colorbar()
-    plt.show()
+    # Uncomment to plot.
+    # field_cutde = u_cutde[:,0].reshape((nx, ny))
+    # field_okada = u_x_okada.reshape((nx, ny))
+    # plt.subplot(1,2,1)
+    # plt.contourf(x_obs_mat, y_obs_mat, field_cutde)
+    # plt.colorbar()
+    # plt.subplot(1,2,2)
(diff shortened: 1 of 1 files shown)


celeri.py
```diff
@@ -2237,7 +2237,7 @@ def test_okada_equals_cutde():
     # Parameters for Okada
     segment_locking_depth = 1.0
     segment_dip = 90
-    segment_length = 0.5
+    segment_length = 1.0
```
tbenthompson, line 2240: Since this length is divided by two below, it should be `1.0` to match with the `[0.5, 0.5]` x range for the triangles.

celeri.py
```diff
@@ -2249,7 +2249,7 @@ def test_okada_equals_cutde():
                 y_obs[i],
                 z_obs[i],
             ],  # (meters) observation point
-            -segment_locking_depth,  # (meters) depth of the fault origin
+            segment_locking_depth,  # (meters) depth of the fault origin
```
tbenthompson, line 2252: Depth is measured in positive downwards for Okada.

celeri.py
```diff
@@ -2271,10 +2271,10 @@ def test_okada_equals_cutde():
     tri_z1 = np.array([0, 0])
     tri_x2 = np.array([0.5, 0.5])
     tri_y2 = np.array([0, 0.0])
-    tri_z2 = np.array([0, 1])
+    tri_z2 = np.array([0, -1])
```
tbenthompson, line 2274: z coordinate should be negative for below the free surface with cutde

celeri.py
```diff
@@ -2290,17 +2290,41 @@ def test_okada_equals_cutde():
...
-    # u_z_cutde[i] = 0
-
-    # TODO: Try grid of observation points
-    # TODO: Try positive z thing again
-    # TODO: Check sign of slip
-    print(disp_mat.shape)
-    print(disp.shape)
-    print(u_cutde.shape)
+
+    # What to do with array dimensions here? We want to compute the
```
tbenthompson, line 2294: Hopefully this helps explain the array dimensions and how you might debug this in the future!

celeri.py
brendanjmeade: I understand this better now.  It looks like I somehow got it right in the segment case.

celeri.py
brendanjmeade: It looks like I also have change this in get_okada_displacements (near line 1040).  Thanks!

brendanjmeade:
Thank you very much for this.  It feels good to have this test and it'll be the foundation for me to implement TDEs everwhere!

PR #57, opened 2021-12-17
Triangular slip constraint development

jploveless:
Currently implemented in a code block in `celeri.ipynb`. Need to move to a function within `celeri.py`. Currently only works for setting any edge combination to creeping. 

- Modified `mesh_parameters.json` to force creep on top, bottom, and sides, just to test method. Tested with [0, 1, 0], [1, 1, 0], and [1, 1, 1] and all work
- Modified general smoothing solution. Previously, smoothing matrix was multiplied by smoothing weight in the operator construction, but smoothing weight should instead go into the weighting vector. I'm curious about weighting vector construction (raw data uncertainties used, rather than reciprocal of their squares), and clearly this implementation of smoothing yields wacky TDE slip rate values 
- Added two new utility functions to `celeri.py`: `get_2component_index` and `get_3component_index`. These general index arrays multiple components given indices of an object. For example, if we want to find the index into a slip component matrix for elements [2, 7, 10], `get_2component_index` will return [3, 4, 13, 14, 19, 20].

### DIFF
--- a/celeri/__init__.py
+++ b/celeri/__init__.py
@@ -21,6 +21,8 @@ from .celeri import (
     get_strain_rate_centroid_operator,
     get_mogi_operator,
     get_keep_index_12,
+    get_2component_index,
+    get_3component_index,
     interleave2,
     post_process_estimation,
     plot_meshes,
--- a/celeri/celeri.py
+++ b/celeri/celeri.py
@@ -2306,7 +2306,6 @@ def get_2component_index(indices: np.array) -> np.array:
     idx = np.sort(np.append(2 * (indices + 1) - 2, 2 * (indices + 1) - 1))
     return idx
 
-
 def get_3component_index(indices: np.array) -> np.array:
     """Returns indices into 3-component array, where each entry of input array 
     corresponds to three entries in the 3-component array
@@ -2324,7 +2323,6 @@ def get_3component_index(indices: np.array) -> np.array:
     """
     idx = np.sort(np.append(3 * (indices + 1) - 3, (3 * (indices + 1) - 2, 3 * (indices + 1) - 1)))
     return idx
-
 
 def post_process_estimation(
     estimation: Dict, operators: Dict, station: pd.DataFrame, index: Dict
--- a/celeri/celeri.py
+++ b/celeri/celeri.py
@@ -201,7 +201,7 @@ def read_data(command_file_name):
     else:
         sar = pd.read_csv(os.path.join(base_folder, command.sar_file_name))
         sar = sar.loc[:, ~sar.columns.str.match("Unnamed")]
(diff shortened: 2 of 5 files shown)


brendanjmeade:
Before merging this, I wonder if we could Zoom quickly.  The reason I'm asking is that it looks like this PR had led to some changes that are breaking the tests, and they seem to be related to the number of returned values expected from some of the file read in commands.   I'd like to hunt these down and get them fixed up!  @jploveless Let me know if you have time on Friday!

> * Modified `mesh_parameters.json` to force creep on top, bottom, and sides, just to test method. Tested with [0, 1, 0], [1, 1, 0], and [1, 1, 1] and all work

While we're here should we move from [A, B, C] specificatn in the .json file (and elsewhere) to something more explicit.  Maybe "top_mesh_bc", "bottom_mesh_bc", and "edge_mesh_bc"?


> * Modified general smoothing solution. Previously, smoothing matrix was multiplied by smoothing weight in the operator construction, but smoothing weight should instead go into the weighting vector. I'm curious about weighting vector construction (raw data uncertainties used, rather than reciprocal of their squares), and clearly this implementation of smoothing yields wacky TDE slip rate values

You are correct!  I just threw in dummy numbers for convenience and never updated them.  Thanks!


> * Added two new utility functions to `celeri.py`: `get_2component_index` and `get_3component_index`. These general index arrays multiple components given indices of an object. For example, if we want to find the index into a slip component matrix for elements [2, 7, 10], `get_2component_index` will return [3, 4, 13, 14, 19, 20].

The latter seems similar to "get_keep_index_12".  Should we unify these?

brendanjmeade:
@jploveless This is a huge and fantastic PR!

brendanjmeade:
@jploveless this runs for me locally!  Quick question.  I see where "top_slip_rate_constraint" etc. get read in and assigned as elements of the meshes list of dictionaries.  What I can't seem to find is where they are later used to apply the constraints.  In the PR version all three Cascadia BCs are set to one and the solution, at least for me, still look unconstrained. Thoughts?

Screen Shot 2021-12-17 at 11 19 32 PM

jploveless:
Currently it’s only in the notebook as a distinct section. I need to
functionalize it in celeri.py but was unsure about updating
operators.meshes. I will play around with this as maybe it’s more
complicated than I’m making it. It works for me as long as I run the code
block within the notebook before doing the estimation block, but I agree
that they’re not getting applied in your figure and that’s something we
need to sort out!

brendanjmeade:
@jploveless Thanks and I see it now!  Some how VSCODE was showing me an older cached version!  I'll update the uncertainty weighting and hope that we can start to recover some classical looking coupling distributions on Cascadia!

PR #58, opened 2021-12-20
Moved TDE constraints to function within celeri.py

jploveless:
And added a call to `celeri.get_tde_slip_rate_constraints` in the non-elastic section of `celeri.ipynb`


brendanjmeade:
Some of the tests have a mamba fail but it doesn't seem general so I'm ignoring it for now.

PR #63, opened 2022-01-07
[do not merge] This went nowhere

tbenthompson:
Do not merge. You can see in the estimating_tde_slip_compact.ipynb notebook what I tried as far as creating a boundary basis of the boundary node set. It's a weird thing to do and it clearly didn't work. But I figured I'd share anyway. I'm assuming the idea is fundamentally flawed but maybe I just have a bug in the code.


brendanjmeade:
Thanks for sharing, and this is really creative!  I won't merge but very cool thinking!

PR #75, opened 2022-02-05
H-matrix solving works. My work is done.

tbenthompson:
Some notes on this:
- I changed the H matrices so that all the internal blocks are 2-dimensional instead of the old 4-dimensional internal matrices.
- I added the tranpose_dot function to HMatrix for when we need to do `v @ A`.
- You'll need to re-run the pickle.dump from celeri_hmatrix.ipynb.
- Then, hmat_dev.ipynb has the fun stuff!
- The main three challenges were getting the weighting right, setting up the column norm preconditioner and then getting the sparse matvec and rmatvec functions correct.


tbenthompson:
(The notes are in the context of my discussion with Brendan earlier, sorry Jack if they don't make a ton of sense)

PR #76, opened 2022-02-09
Fixed indexing but partials calculations have empty columns?

jploveless:
The inversion fails with a singular matrix error. I assume this is due to some columns being all zeros, and this seems to be the case from some of the troubleshooting I placed into the bottom of the solver code block, but I can't figure it out. Looks like there are zero columns corresponding to the first few columns of TDE partials (63, 64, 65…) but then when I try to check that matrix of partials itself, it's not reporting any zero columns, so clearly my Python inexperience is causing some issues here. But, if the TDE partials indeed contain some all-zero columns, I'm not sure how to fix that. 

Any ideas?


brendanjmeade:
I'm merging this.

@jploveless From your description this sounds like a matrix assembly issue likely focused on `estimation.operator`?

brendanjmeade:
It looks like most of the `estimation.operators` matrix is empty right now, so that's something to go on!
spy(estimation operator)

jploveless:
Okay, I think I have some indexing suspicions about inserting the TDE
partials into estimation.operators, and I will try to check that out soon.

PR #78, opened 2022-02-11
Removed meshes item from operators, index dicts

jploveless:
PR for edited `operators` and `index` dicts when referring to mesh-related variables


brendanjmeade:
I'll work to fix the tests.

Issue #84, opened 2022-02-15
Stations being assigned to the wrong blocks!!!


brendanjmeade:
For the Japan model, GPS stations are being assigned to the wrong blocks in some cases.  This is visually documented in the file: `notebooks/celeri_direct_dense_japan_debug.ipynb` in the commit: https://github.com/brendanjmeade/celeri/commit/e4bd6427544280a8b27ac721772b1835e057f1a8

_Run this notebook and the last two cells will produce figures that show the station misassignment._

Some notes/observations:

- The block polygons appear to be closed and labeled properly for the Japan model!
- The station mislabeling only appears to be a problem for blocks 1, 2, 3, 4, and 6 ("rest of the world" block).
- Block 6 is the old rest of the world block
- Most of the blocks with stations assigned to the wrong block touch the "rest of the world" block.  Maybe not all, though, as the Philippine Sea plate seems to be fine?

@tbenthompson @jploveless This seems to indicate that there is an issue related to assigning stations to the correct blocks after they have been properly closed.  Perhaps in:
https://github.com/brendanjmeade/celeri/blob/e4bd6427544280a8b27ac721772b1835e057f1a8/celeri/celeri_closure.py#L241

Maybe the "rest of the world" block is causing this problem and we should just get rid of it by introducing some segments that make a complete circuit around the globe (e.g. near the equator)?

tbenthompson:
I'll take a look this week. Probably tomorrow if I have time. I'll aim to
reduce it down to a reproducible bug with a single polygon and single point
and see if that helps clarify!

brendanjmeade:
Thanks @tbenthompson!  If you like, I'm available on Zoom tomorrow after 1:30.  Also, if the best answer is to get rid of the "rest of the world"/"exterior" block that would be a great solution too!

jploveless:
This does indeed seem like an issue with the exterior block. In the Japan model, it is labeled as block 6, and when it is circulated to determine which points (stations) lie within, it is overprinting the block labels of stations assigned to blocks with labels ≤ 5. In Blocks, we somehow had a way of circulating around the exterior block in the opposite direction, so that it truly encompassed the rest of the world, rather than encompassing all of the block geometry that lies within. 

[...]

PR #86, opened 2022-02-16
Added a more sophisticated mechanism for identifying a good interior point

tbenthompson:
Followup to https://github.com/brendanjmeade/celeri/issues/84:

This PR adds a more sophicasted  mechanism for identifying a good interior point:
- we loop over segments and choose an interior point that is 25% of the segment length away from the polygon towards the interior. 
- if the line segment from the midpoint of the segment under consideration to the proposed interior point intersects any other polygon edges, then that interior point is rejected. 

I did some other small things:
- I commented out the end2end test since it's super slow and annoying and has been failing for over a month anyway.
- I moved the tests at the bottom of celeri_closure.py into a new test_closure.py file and added a couple new tests that verify the behavior in this PR.


tbenthompson:
The resulting choices of interior points look pretty nice.

image

tbenthompson:
With the latest commit, the rogue point is now correctly included inside block 4:
image

brendanjmeade:
@tbenthompson and @jploveless Thanks for the discussion and fix.  I had a very good time making that table last night.  I can confirm that things appear to be getting labeled properly when I run things.  It's great that we were all able to replicate and see the issue.  Thanks both!

PR #88, opened 2022-02-25
Fixed multiple mesh plotting

jploveless:
Previously in `celeri_vis.py`, mesh geometries were looped across, but the subset of slip component values was not extracted, so it would fill the triangles using values `0:meshes[i].n_tri` for all meshes. Now the subset is taken and slip rates look good! Still need to figure out edge constraints, as those don't seem to be respected. If anything it seems like no-slip conditions are being applied regardless of settings.


brendanjmeade:
This is fantastic @jploveless!  It feels so good to see these distributions again.

PR #122, opened 2024-12-16
Coupling02

brendanjmeade:
Iterative quadratic approach to estimating slip distributions within coupling bounds.  The bounds depend on the kinematic velocities resolved onto 3D mesh surfaces.  Some changes to celeri.py here (including unrelated reordering of logging information and hdf output of results).  The core implementation is done for a three mesh (+ one CMI mesh) Japan example and is found in: celeri_japan_qp_cmi_coupling.ipynb


jploveless:
What a tour de force! This is a really nice approach, and I feel like it's written in a really readable form. I'm able to run the entire notebook, minus the fancy figures (convergence evolution plots and nice contour plots; I saw them in the GitHub notebook but I wasn't able to remake them during the run, presumably because of some plot/render setting on my own install, so I'm not super worried about it (but dismayed that I can't make the figures, because they're gorgeous!)). 

I think there might still be some logic to implement when adding the approach to `celeri.py` in terms of choosing the solution method (for example, does this still count as `"inversion_type": "standard"` (in the `.command` file), and everything will need to be wrapped in a `if meshes[mesh_idx].qp_mesh_tde_bound_coupling == 1` statement). Those feel straightforward and I'd be happy to chip away at them, though I do wonder if it's an all-or-none solution method (i.e. is it possible to do an unbounded estimate for some meshes but not all? If it's better practice to say all-or-none, maybe we should make the flag exist in the `.command` file (which could be a good opportunity to clean up some of the overlapping references to `inversion_type`, `solution_method`, and `solve_type`). 

This is amazing!

brendanjmeade:
@jploveless Thanks for the comments.   

- Right now, the solution is all or none.  It's certainly possible to modify the logic here.
- I think there are a lot of options in terms of how we specify what type of model run we want from a command file and/or the command line.  Currently, I'm not focused on this because it runs fast and the custom plotting in the notebook takes the most time, so running directly in a notebook is efficient for now.  Good to think ahead!

I'll merge the PR.  Thanks!

PR #124, opened 2025-03-19
Check test status on different Python versions

maresb:
I want to determine the min and max supported Python versions


maresb:
Modern versions of okada-wrapper require Python >=3.8

PR #125, opened 2025-03-19
Trigger pytest workflow on PR

maresb:
When trying to run tests in #124 I noticed that the pytest workflow doesn't run on PRs. This fixes that.


maresb:
Blocked on #126

PR #134, opened 2025-03-23
Fix test CI

maresb:


maresb:
We should drop Python 3.7 because there's no corresponding cutde build.

Also, cutde is blocking support for linux-aarch64 (Docker on Apple Silicon) and Python 3.13. A feedstock PR is incoming thanks to https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/7204. I'll need to get on the feedstock as per https://github.com/conda-forge/cutde-feedstock/pull/14 to push things through.

maresb:
micromamba is taking suspiciously long. Should probably just go directly to pixi...

maresb:
Another dependency observation: spherical-geometry is not available for python 3.8 on win-64 or osx-arm64. For this reason, we should probably require at least Python 3.9 in our first release.

maresb:
@brendanjmeade, now that the tests are green, we have basic guardrails to know that the package isn't totally screwed up. After merging this, I think we are clear to attempt the first PyPI release.

PR #135, opened 2025-03-26
Replace okada_wrapper with cutde

### LINKED ISSUE #70: Replace all Okada calls with two TDEs instead

maresb:
Closes #70

TODO: vectorize so that it's not horrendously slow.

Currently I reimplemented a drop-in replacement that computes one observation at a time, but there's a lot of overhead from that.


maresb:
Need to perform an analysis of the errors to ensure they're within expected tolerances. Also need to write test cases to demonstrate that, within those tolerances, okada_wrapper = cutde.

brendanjmeade:
@maresb This is such a huge step forward!  I'm going to have to get my "Okada" tattoo removed ;)

maresb:
@brendanjmeade, on the Japan test, there was just one single major discrepancy between okada_wrapper and cutde.

I'm fairly certain that in the okada_wrapper version, the station is being lightsabered by a numerical artifact, which I understand is fairly typical. In contrast, the cutde version is very stable. I'll do more tests when I return, but this looks very promising to me for the cutde version.

For the lightsaber, see the images at the bottom of [the notebook I just pushed into this branch](https://github.com/brendanjmeade/celeri/blob/ae7a80ce31c8005a965f602be5a69a429a19203a/notebooks/okada_wrapper_glitch.ipynb). The x and y axes are perturbations of the station lon/lat in units of 0.02 degrees. The value plotted is the up-component of displacement from tensile slip, minus the cutde value with no perturbation.

Top (glitchy) is okada_wrapper, bottom (continuous) is cutde:
image

Away from the line the values agree.

brendanjmeade:
@maresb Thanks for doing and sharing all of this!

> I'm fairly certain that in the okada_wrapper version, the station is being lightsabered by a numerical artifact, which I understand is fairly typical. In contrast, the cutde version is very stable. I'll do more tests when I return, but this looks very promising to me for the cutde version.

"Lightsabered" is now permanently in my vocabulary!  Okada's code is original but doesn't have the types of corrections for along ray issues like this that cutde does (see below).

> Side note: While tame compared to the pathologies you're used to struggling with, I recently realized that the cutde triangulation will unfortunately introduce an artificial singularity along the diagonal that we choose to triangulate along. Probably we don't care, but as a mathematical perfectionist it annoys me. (If we really wanted to, we could intelligently switch diagonals, but then the rectangle midpoint is still always singular; to cover that case we could use a 3-triangle triangulation, but I won't go there.)

I have good news for you.  The cutde is built on the Nikhoo and Walter code, and they've done a lot of those corrections (see "[Triangular dislocation: an analytical, artefact-free solution](https://academic.oup.com/gji/article/201/2/1119/572006)").  By way of honesty it's my old matlab code for this that introduced a lot of the artefacts.


> 1. repeat this analysis for north america data
> 2. systematically check edge cases like negative width and height

These are fantastic goals.

I'm guessing the speed will be nearly identical?

Nice!

maresb:
Here's the corresponding scatter plot for western north America. As a refresher, lower is better, and the line at the top doesn't matter as long as it's to the left.

image

Most of the anomalies seem to be caused by the tensile-up components. Here's what it looks like when those are masked:

image

And as one could have guessed, here's the lightsaber for the tensile-up component of the leading point in the first scatter:

image

maresb:
Here is cutde vs cutde comparing different triangulations. I define three triangulations of the rectangle: `/`, `\`, `V`. I describe them in detail here (scroll down in this pane to admire my ASCII art!): https://github.com/brendanjmeade/celeri/blob/f185e1e9d5c9e9533f3b3f87fce669f4134095e8/celeri/celeri_util.py#L47-L73

Here's the scatter plot to be compared with those above (difference is `V` vs `/`):

image

Perhaps more insightful is the absolute error, because it removes the slope, which is effectively an artifact of the "relative" parameterization:

image

No more slope, and the absolute error is always less than $10^{-11}$.

In more detail, I plot the noisiness of the lat/lon dependence of the north component of displacement per unit strike slip between station 0 and segment 116:

`/` triangulation:
image

`\` triangulation:
image

`V` triangulation:
image

Things to note:

* As expected, given that the magnitude of the displacement component is about $10^{-9}$, and the absolute error is less than $10^{-11}$, so the error is less than 1%.
* As expected, the `/` and `\` triangulations are less noisy, presumably because the `V` triangulation has more opposing edges, and with longer total length.

Based on this, I'm going to declare cutde the clear winner.

maresb:
@brendanjmeade, I originally had plans to make an exact drop-in lightsaber-free `cutde` replacement for the displacement component of `okada_wrapper.dc3dwrapper`, so that the following are exactly equivalent up to computational artifacts:

```python
_, u, _ = okada_wrapper.dc3dwrapper(
    alpha, xo, depth, dip, strike_width, dip_width, dislocation
)
u = dc3dwrapper_cutde_disp(
    alpha, xo, depth, dip, strike_width, dip_width, dislocation
)
```

One issue I noticed is that when `dip_width[0] > dip_width[1]`, I need to flip the strike and tensile components (but not dip) of `dislocation` in order to get `cutde` to reproduce the expected results. And this case actually occurs in the Japan and WNA segments, which is why I noticed it.

**My question for you**: would you like me to further analyze and reproduce the atypical cases in `okada_wrapper.dc3dwrapper` (such as when `strike_width[0] > strike_width[1]` and when both strike and dip widths are backwards)? Or would you prefer that I skip this and get straight to the vectorization and the unwinding the now-unnecessary coordinate transformations? (With `okada_wrapper.dc3dwrapper` fresh in my head, doing it now would be much easier for me than if we were to postpone till later.)

brendanjmeade:
@maresb 

> Based on this, I'm going to declare cutde the clear winner.

Absolutely beautiful documentation and argument.  The ASCII art is honestly clear and helpful.  All of this makes sense and I agree with all of it.  

This is going to put the entire project on a much better foundation and enable us to move forward with confidence!

> One issue I noticed is that when dip_width[0] > dip_width[1], I need to flip the strike and tensile components (but not dip) of dislocation in order to get cutde to reproduce the expected results. And this case actually occurs in the Japan and WNA segments, which is why I noticed it.

I'm intrigued to learn that we face this case.  My expectation would be that we would not.  Can you point me to a segment where this is the case?  I'd like to look at this and understand why this is happening.

I **definitely** think it is worth it for you to continue hunting down and understanding these cases.  Again, this will give us confidence.  Much appreciated and please let me know if there's anything I can do to help!

brendanjmeade:
@maresb Thanks for sending this!

I can certainly verify that there are a bunch of non-zero `burial_depths`.  I have no idea how these got there because burial depth hasn't been used in probably a decade, and I don't see any realistic future use for them.

The one reason to keep them in the segment files for now is that the web-based UI for constructing segment files (https://github.com/brendanjmeade/celeri_ui) assumes that the `burial_depth` (and related) columns exist.  I'm not confident that I can remove that from the celeri_ui codebase without breaking it.  Sounds like a ChatGPT yolo experiment!

For now, I recommend something like the following:
- If `burial_depth` is in a segment file, either set all values to zero on read in or ignore it.
- In code, either set burial_depth to zero everywhere, or just eliminate it.

Questions:
- @maresb @aseyboldt Want me to modify the `*segment.csv` files to set all burial depths to zero?  Or would this break testing at this time?
- @jploveless Ready to nuke all `burial_depth`s?

jploveless:
I don't think I've used a nonzero burial depth in my career, so I am okay with nuking them! 

@maresb Thank you for such clear documentation of the Okada to cutde transition! It's really cool to see how well the replacement works, and I've learned a lot from the special cases you've presented. I'm especially glad that the sign issues can be resolved by simply getting rid of `burial_depth`!

PR #138, opened 2025-04-10
Formatting and linting

maresb:
Formatting reduces entropy and linting automatically catches various mistakes, sloppiness, and outdated Python paradigms.

As an example, we get rid of unused imports, like the one preventing celeri from working on Python 3.13.

Disabled for the moment is formatting and linting of notebooks. Also, several rules were disabled for follow-up in order to make this PR more manageable.

PR #139, opened 2025-04-11
Use pathlib

maresb:
Builds on #138. Once that's merged, this PR will become short.

We should use the `pathlib` API for doing path manipulations since it's more reliable and standardized.


aseyboldt:
I think this needs a rebase now that #138 is merged?

PR #141, opened 2025-04-11
Explicitly check if lockfile needs to be updated

maresb:
Before, if the lockfile wasn't up-to-date, the error message was somewhat vague.

This just adds a separate step where we check the pixi lockfile, so if the lockfile isn't up-to-date, then the explicit `Check if pixi lockfile is up-to-date` step will fail, hopefully making it easier to understand what's wrong.


maresb:
Reported the pixi cleanup failure upstream to https://github.com/prefix-dev/setup-pixi/issues/192

maresb:
@brendanjmeade, ready to merge

PR #136, opened 2025-04-04
WIP: Experiment with convex optimization

aseyboldt:
Not ready for review yet, just for reference and to share with @maresb

This includes some unfinished code to iteratively optimize convex subproblems in the japan example.

PR #145, opened 2025-04-16
Add new iterative convex optimizer

aseyboldt:
Cleaned-up version of #136.

This adds a new module `optimize`, that implements a modification of the sequential convex optimization approach, which seems to converge much faster.

I didn't touch the notebooks with the original implementation, as we might want to compare them a bit more.

@brendanjmeade This also includes a benchmarking notebook. I'd be curios what results you get on your machine (and also with gurobi) in those benchmarks.


aseyboldt:
@maresb I also added some version constraints in the pixi file. Or did you leave those out on purpose?

maresb:
My omission of version constraints if primarily philosophical because I don't see any purpose. (If you want a known working solution, that's what the lockfile is for, and semver guarantees are only theoretical rather than practical.) Thus my strategy is to use `*` unless there's a specific reason. And then I always try and leave an explanatory comment with GitHub link. This system keeps entropy low. (Otherwise I look at it and wonder if the constraint is real or fake.)

In case you want to go into an hour's-long rabbit hole, there's [this article](https://iscinumpy.dev/post/bound-version-constraints/) which I found really insightful, but unless you have a weird obsession with package management like me, there are probably much better uses of your time.

brendanjmeade:
Running this and hitting the error below:

Maybe I don't have cvxpy / clarabel installed?  I didn't see them as imports in the notebook so I'm not sure.

Guidance welcome!

```
---------------------------------------------------------------------------
PanicException                            Traceback (most recent call last)
Cell In[4], line 10
      8 kwargs = dict(zip(setting_lists.keys(), args))
      9 solver, solve_args = kwargs.pop("solver")
---> 10 out = celeri.optimize.benchmark_solve(
     11     problem=problem,
     12     solver=solver,
     13     solve_kwargs=solve_args,
     14     **kwargs,
     15 )
     16 out["solver_path"] = str((solver, solve_args))
     17 out["config"] = str(
     18     (
     19         solver,
   (...)
     23     )
     24 )

File ~/Desktop/celeri/celeri/optimize.py:1261, in benchmark_solve(problem, with_limits, objective, rescale_parameters, rescale_constraints, solver, solve_kwargs)
   1259 start = time.time()
   1260 try:
-> 1261     _custom_solve(
   1262         minimizer.cp_problem,
   1263         solver=solver,
   1264         objective=objective,
   1265         **default_solve_kwargs,
   1266     )
   1267 except Exception as e:
   1268     success = False

File ~/Desktop/celeri/celeri/optimize.py:1120, in _custom_solve(problem, solver, objective, **kwargs)
   1118     _custom_cvxopt_solve(problem, **kwargs)
   1119 else:
-> 1120     problem.solve(solver=solver, **kwargs)

File ~/mambaforge/envs/celeri/lib/python3.10/site-packages/cvxpy/problems/problem.py:503, in Problem.solve(self, *args, **kwargs)
    501 else:
    502     solve_func = Problem._solve
--> 503 return solve_func(self, *args, **kwargs)

File ~/mambaforge/envs/celeri/lib/python3.10/site-packages/cvxpy/problems/problem.py:1086, in Problem._solve(self, solver, warm_start, verbose, gp, qcp, requires_grad, enforce_dpp, ignore_dpp, canon_backend, **kwargs)
   1084 if solver_verbose and (not verbose):
   1085     print(_NUM_SOLVER_STR)
-> 1086 solution = solving_chain.solve_via_data(
   1087     self, data, warm_start, solver_verbose, kwargs)
   1088 end = time.time()
   1089 self._solve_time = end - start

File ~/mambaforge/envs/celeri/lib/python3.10/site-packages/cvxpy/reductions/solvers/solving_chain.py:473, in SolvingChain.solve_via_data(self, problem, data, warm_start, verbose, solver_opts)
    437 def solve_via_data(self, problem, data, warm_start: bool = False, verbose: bool = False,
    438                    solver_opts={}):
    439     """Solves the problem using the data output by the an apply invocation.
    440 
    441     The semantics are:
   (...)
    471         a Solution object.
    472     """
--> 473     return self.solver.solve_via_data(data, warm_start, verbose,
    474                                       solver_opts, problem._solver_cache)

File ~/mambaforge/envs/celeri/lib/python3.10/site-packages/cvxpy/reductions/solvers/conic_solvers/clarabel_conif.py:330, in CLARABEL.solve_via_data(self, data, warm_start, verbose, solver_opts, solver_cache)
    326     _results = _solver.solve()
    328     return _results, _results.status
--> 330 results, status = solve(solver_opts)
    332 if solver_cache is not None and self.STATUS_MAP[str(status)]:
    333     solver_cache[self.name()] = results

File ~/mambaforge/envs/celeri/lib/python3.10/site-packages/cvxpy/reductions/solvers/conic_solvers/clarabel_conif.py:325, in CLARABEL.solve_via_data.<locals>.solve(_solver_opts)
    322 def solve(_solver_opts):
    324     _settings = CLARABEL.parse_solver_opts(verbose, _solver_opts)
--> 325     _solver = clarabel.DefaultSolver(P, c, A, b, cones, _settings)
    326     _results = _solver.solve()
    328     return _results, _results.status

PanicException: Unrecognized LDL solver type
```

aseyboldt:
Ah, it seems that the conda-forge clarabel for some reason doesn't come with support for faer. I'll see if I can figure out why...
In the meantime you can just get rid of that setting. So in the config of the solvers you can disable the one that uses faer:

```
    "solver": [
        ("CLARABEL", {"equilibrate_enable": False}),
        # ("CLARABEL", {"equilibrate_enable": False, "direct_solve_method": "faer"}),
        ("CLARABEL", {"equilibrate_enable": True}),
        ("CUSTOM_CVXOPT", {}),
    ],
```

And later on in the whole pipeline as well:

```
%%time
problem = celeri.optimize.build_problem(command_files["north_america"])

solve_kwargs = dict(
    solver="CLARABEL",
    equilibrate_enable=False,
    # direct_solve_method="faer",
    presolve_enable=False,
    warm_start=True,
    ignore_dpp=False,
)

trace_north_america = celeri.optimize.minimize(
    problem,
    verbose=True,
    velocity_upper=110.0,
    velocity_lower=-110.0,
    solve_kwargs=solve_kwargs,
    objective="sum_of_squares",
)
```

aseyboldt:
> My omission of version constraints if primarily philosophical because I don't see any purpose. (If you want a known working solution, that's what the lockfile is for, and semver guarantees are only theoretical rather than practical.) Thus my strategy is to use * unless there's a specific reason. And then I always try and leave an explanatory comment with GitHub link. This system keeps entropy low. (Otherwise I look at it and wonder if the constraint is real or fake.)

Yeah, it is unfortunate that really nobody follows semver in python...

I think at least lower bounds are a good idea. Without those, problems on the packaging side, and we might suddenly end up with some broken ancient version of something?

brendanjmeade:
I've commented out the faer in three places.  The notebook runs for a while.  

When it reaches the line:

```python
now = datetime.datetime.now(datetime.UTC)
```

in the fourth code cell it now throws the following error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[4], line 37
     28 ds = pl.from_dicts(results)
     29 ds = ds.with_columns(
     30     params=ds["params"].map_elements(
     31         lambda x: np.array(x), return_dtype=pl.List(pl.Float64)
   (...)
     35     ),
     36 )
---> 37 now = datetime.datetime.now(datetime.UTC)
     38 filename = f"{now.isoformat(timespec='seconds')}_benchmark_opt_solver.parquet"
     39 ds.write_parquet(filename)

AttributeError: module 'datetime' has no attribute 'UTC'
```

aseyboldt:
`datetime.UTC` was added in python 3.11, so it looks like you are on a different version? So probably not using the pixi setup?
You can just change that line to `now = datetime.datetime.now()` though.

brendanjmeade:
Thanks @aseyboldt !  I'm running it now with `now = datetime.datetime.now()`.  Is it ok to install with the pixi installer now?  I had some vague idea that @maresb was still working on full cutde integration?

aseyboldt:
I think the work on cutde integration was independent of that? But maybe I missed something (@maresb).
It should work for the benchmarking notebook. You can also add gurobi to the environment with `pixi add --pypi gurobipy`. If the license is at one of the standard locations, I *think* it should work automatically then.

aseyboldt:
For each of the datasets we run all difference combinations of solver and model settings.

We have three solvers: clarabel, cvxopt (not exactly what's used in the original optimizer, but I think very close. We might want to double check that still a bit though) and gurobi. We then run these with different settings: With manual column rescaling, constraint rescaling; with column rescaling and no constraint rescaling; no column rescaling and constrain rescaling and no column rescaling and no constraint rescaling. Those are the `config` in different colors. Clarabel also has an extra setting to enable or disable automatic rescaling.

We also run each dataset once with kinematic velocity limits and once without (the row of the plot).

Finally, we change the way we code the optimization problem. (the columns). The first three are mathematically identical, but can behave quite differently numerically. In these three we want to minimize the 2-norm of the residuals, so $\min |Ax - b|_2$.

1. expanded_norm2: We write the objective as $(Ax - b)^T(Ax-b)$ and expand the expression. We end up with $x^T(A^TA)x + 2(b^TA)x$ plus a constant. This was used in the original code.
2. Here, we introduce new variables r with equality constraint r = Ax - b and then minimize $r^Tr$.
3. This is similar to two, but we push the quadratic term to a second order cone. So this is a linear objective with a second order cone constraint.
4. Here, we minimize the 1-norm of the residuals. This is a linear problem with no second order cone constraint.


Let's start with the second, the `objective` plot. This shows the final value of the objective function (the 2-norm of the residuals) at the point the optimizer returned. The different solver settings seem to mostly agree, except for the first objective. Here, we often get quite different results depending on the solver settings. So it seems this is quite unreliable. I *think* the problem is that the objective value the solver sees never really change in the last couple of digits (we dropped the constant term of the original constraint). The solvers stop if the dual and primal objectives are close, and I think this large offset messes that up.

Runtime wise, it looks like gurobi beats all other solvers in the second objective variant. Funnily, it seems to do terribly in the third one... Clarabel also doesn't look bad when we disable the automatic rescaling. That seems to make things worse for some reason.

---

The last couple cells run the full optimizer with multiple iterations. What times do you see there?

aseyboldt:
oh, and can you mail me the paraqet file that is written after the solver runs?

aseyboldt:
Well,that doesn't look too bad :-)

You should also be able to run it with gurobi now:

```python
%%time
problem = celeri.optimize.build_problem(command_files["japan"])

solve_kwargs = dict(
    solver="GUROBI",
    ignore_dpp=True,
)

trace = celeri.optimize.minimize(
    problem,
    verbose=True,
    velocity_upper=110.0,
    velocity_lower=-110.0,
    solve_kwargs=solve_kwargs,
    objective="sum_of_squares",
)
```

for japan, or for north america if you change to the other command file in the first line.

brendanjmeade:
@aseyboldt Great idea!  I've run this now and below I report some run times:

**Japan models**:
1. `celeri_japan_qp_coupling_linear_smoothing.ipynb`: Initial sequential quadratic programming (CVXOPT): 1 min 4 seconds
2. `benchmark_convex_solvers.ipynb`: Linear envelopes (not sure) (Clarabel): 1 min 12 seconds
3. `benchmark_convex_solvers.ipynb`: Linear envelopes (not sure) (Gurobi): 0 min 26 seconds
4. `benchmark_convex_solvers.ipynb`: Linear envelopes (not sure) (CVXOPT): Failed

I tried 4 with
```python
%%time
problem = celeri.optimize.build_problem(command_files["japan"])

solve_kwargs = dict(
    solver="CVXOPT",
    equilibrate_enable=False,
    # direct_solve_method="faer",
    presolve_enable=False,
    warm_start=True,
    ignore_dpp=True,
)

trace_japan = celeri.optimize.minimize(
    problem,
    verbose=True,
    velocity_upper=110.0,
    velocity_lower=-110.0,
    solve_kwargs=solve_kwargs,
    objective="sum_of_squares",
)
```


**Western North America models**:
1. `celeri_japan_qp_coupling_linear_smoothing.ipynb`: Initial sequential quadratic programming (CVXOPT): ~3 min
2. `benchmark_convex_solvers.ipynb`: Linear envelopes (not sure) (Clarabel): 1 min 3 seconds
3. `benchmark_convex_solvers.ipynb`: Linear envelopes (not sure) (Gurobi): 0 min 25 seconds

The first question I have is a point of clarification: is the minimization approach applied in `benchmark_convex_solver.ipynb` the linear envelop approach that you developed, or the sequential quadratic programming approach that I developed?

**Some potential points of interest**:
- The initial sequential quadratic programming approach appears to have approximately the same runtime for the Japan case, but is ~3x slower for the previously troublesome Western North America case.  I very much like that we have a reasonable convergence path for North America!

- For these two cases, Gurobi is about 2x faster than Clarabel.  I'm probably willing to live with a factor of 2 to keep things open source!

- For the Japan case, the initial sequential quadratic programming approach has approximately the same runtime as the Clarabel linear envelop approach.  The sequential quadratic programming case has 190 iterations with CVXOPT quadratic programming solve each, and the linear envelopes approach has 13 iterations.  Assuming the time spent in matrix assembly is negligible, this suggests that each iteration is approximately 10x faster in CVXOPT as compared to Clarabel.  Admittedly, they aren't solving the same problem, but they should be close to the same size. 
    - Does this align or differ with how you think about the solution speed in these two cases?
    - Are you using spare or dense constraint matrices?  I'm using dense in the sequential quadratic programming approach, and that's approximately a 100x faster solve than with sparse constraint matrices.  If using sparse matrices, can you try with dense?

aseyboldt:
The optimization at the end is what we discussed in the last meeting.
I think both methods are sequential quadratic programming approaches, and I'm not sure if I'd call the new one an envelope approach (it only uses a true convex envelope in the first iteration, later we are cutting away parts of the space as well). But I guess we should come up with some names to distinguish them.

The `CUSTOM_CVXOPT` solver method is using dense matrices. If I use sparse once I also just get horrible performance with cvxopt. But I think that is mostly because as far as I can tell it is using an iterative solver without preconditioner, and that usually isn't fast (at least for small or medium sized problems as ours). Clarabel uses a direct sparse solver (a custom implementation or faer if you enable that setting), and doesn't even support a dense solver. Maybe that would be faster here, but I don't think it would be all that much faster. Our matrices aren't as sparse as I would like them to be, but they aren't exactly dense either. And the kkt system that it actually solves is even more sparse.

I'm also not sure it really makes too much sense to compare the performance of cvxopt and clarabel. The expanded_norm2 formulation of the problem seems quite unstable with both clarabel and cvxopt, so I think we probably shouldn't be using that anyway. And from some small experiments I did with CVXOPT and the second problem formulation it seems that it performs *much* worse than clarabel in that setting (both with dense and sparse matrices). And if we increase the size of the problem, sparse solvers should get better compared to dense solvers anyway.

maresb:
Hey @brendanjmeade and @aseyboldt, just skimming this really quickly to make a few short comments, sorry if I missed some context:

You're all good to use `pixi` because `pixi` projects handle both Conda and PyPI dependencies. What's still pending is a conda-forge recipe for celeri, and that indeed requires all dependencies to be on conda-forge (and nothing from PyPI), and thus switching to cutde is a prerequisite because okada_wrapper has no license and can't be added to conda-forge.

My 2¢ regarding lower pins, they can't hurt in the `pixi.toml`. Under normal circumstances the solver will generally pick the latest versions, so they also won't help... (unless a conda-forge maintainer adds a dependency constraint that's non-monotonic across versions, in which case pixi's solver must arbitrarily choose which package to downgrade. That doesn't happen often, and when it does I try to fix it in conda-forge ASAP). The corresponding question for `pyproject.toml` is a different story because that would actually limit the environments in which celeri could be installed without a concrete reason.

brendanjmeade:
Thanks @aseyboldt!

I think something I'd like to understand is how your method works on the real Japan and Western North America models if we use dense matrices and the CVXOPT solver.  As I mentioned previously in the thread, I tried (and admittedly ill-informed) a solve with CVXOPT and failed.

> I think both methods are sequential quadratic programming approaches, and I'm not sure if I'd call the new one an envelope approach (it only uses a true convex envelope in the first iteration, later we are cutting away parts of the space as well). But I guess we should come up with some names to distinguish them.

Happy to iterate on a name!


> And if we increase the size of the problem, sparse solvers should get better compared to dense solvers anyway.

I agree, and I'd still like to understand if there is a dense CVXOPT version that might run very quickly for problems like the Japan and Western North America ones that we have now.  My partially informed perspective is that this might be a very good combination of QP solve speed and your algorithm, which requires far fewer iterations.

aseyboldt:
I can't test it right now, but you can set the solver to CUSTOM_CVXOPT to
get the dense solver. You also need to delete a couple of solver options
that don't apply to CVXOPT, and set the objective to `expaned_norm2`. But I
don't think that will work. Last time I tried (that was with slightly
different constraints though) the solver would sometimes fail or just
return incorrect ninima, especially in the north America example.

brendanjmeade:
Thanks for the suggestion.  I just tried:

```python
%%time
problem = celeri.optimize.build_problem(command_files["japan"])

solve_kwargs = dict(
    solver="CUSTOM_CVXOPT",
    ignore_dpp=True,
)

trace = celeri.optimize.minimize(
    problem,
    verbose=True,
    velocity_upper=110.0,
    velocity_lower=-110.0,
    solve_kwargs=solve_kwargs,
    objective="expanded_norm2",
)
```

and got the error below:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File <timed exec>:8

File ~/Desktop/celeri/celeri/optimize.py:1184, in minimize(problem, velocity_upper, velocity_lower, max_iter, smooth_kinematic, solve_kwargs, reduction_factor, verbose, rescale_parameters, rescale_constraints, objective)
   1181 solver = default_solve_kwargs.pop("solver")
   1183 for _num_iter in range(max_iter):
-> 1184     _custom_solve(
   1185         minimizer.cp_problem,
   1186         solver=solver,
   1187         objective=objective,
   1188         **default_solve_kwargs,
   1189     )
   1190     trace.store_current()
   1191     if verbose:

File ~/Desktop/celeri/celeri/optimize.py:1118, in _custom_solve(problem, solver, objective, **kwargs)
   1114     if objective != "expanded_norm2":
   1115         raise ValueError(
   1116             "CUSTOM_CVXOPT solver only supports expanded_norm2 objective"
   1117         )
-> 1118     _custom_cvxopt_solve(problem, **kwargs)
   1119 else:
   1120     problem.solve(solver=solver, **kwargs)

File ~/Desktop/celeri/celeri/optimize.py:1054, in _custom_cvxopt_solve(problem, **kwargs)
   1052 # Check that ignore_dpp is the only key
   1053 if len(kwargs) > 1:
-> 1054     raise ValueError("Only 'ignore_dpp' is allowed as a keyword argument.")
   1055 elif len(kwargs) == 0:
   1056     pass

ValueError: Only 'ignore_dpp' is allowed as a keyword argument.
```

I tried commenting out the ignore_dpp keyword argument line in the assignment of the `solve_kwargs` dict and it returned the same error.

Let's consider this next week.

Hope you have a good weekend!

PR #146, opened 2025-04-17
Remove upper pixi version limits

aseyboldt:
As mentioned in #145, this removes the upper version limits for dependencies in pixi.toml


maresb:
Thanks!

PR #148, opened 2025-04-26
Draft refactor addict and split celeri.py

aseyboldt:
@brendanjmeade What would you think about refactoring along these lines? It isn't finished and doesn't fully work yet, as I wanted to get your thoughts before I spend a lot of time working on this.

The idea is to split celeri.py into several files, for instance one for mesh, one for the config (previously called command), one for constants etc...

We can then turn the addicts that store information about meshes and so on into dataclasses. I think that would make things quite a bit more explicit and easier to understand the code.

I accidentally based this refactor or #147, so you can ignore the first two commits for the purpose of this PR.


maresb:
Thanks @aseyboldt!!!

If we get the tests to run it may be worth the effort for me to rebase my other PRs over this.

aseyboldt:
I fixed the tests. Had to add default values to some columns in the `segment` dataframe, as the new code computes some operators the last version didn't.
Maybe it would be a good idea to use pydantic for the global and mesh config, and also validate the dataframes a bit more.

The py39 tests are failing because I used a bit of py310 syntax. @brendanjmeade do we still need py39 compatibility? The scientific python spec [here](https://scientific-python.org/specs/spec-0000/#support-window) suggests we should already require py311 by now (but I see a lot of py310 traffic in various projects still, so not sure about dropping that too).

maresb:
I'm suspecting that there could be users who have old Python 3.9 environments. Thus I'm aiming for a first stable release including 3.9 support before going all-out with the linting rules and updating Python syntax. But maybe we just don't care about this.

I have a lot of stuff in progress e.g. #139 and #135, and I think it may be easier if we could hold off for now on splitting `celeri.py`. Would it be a lot of work to revert that for now?

aseyboldt:
I think most of #139 is on HEAD already? If so, I think both are relatively small, so we should be able to merge in any order, and I think we won't have to do too difficult merging.

aseyboldt:
I'm working through the notebooks now, fixing up the API changes, and making sure the results don't change.

## List of notebooks
- [x] `doc_tde_slip_imaging`
- [x] `doc_resolution_test`
- [x] `doc_pygmt`
- [x] `doc_matplotlib_vis`
- [x] `doc_map_projections`
- [x] `doc_linear_operator_structure` (doen't contain code)
- [x] `doc_benchmark_convex_solvers`
- [ ] `doc_matplotlib_vis_widget`: depends on `read_run`, which still needs to be adapted to the refactor.
- [ ] `doc_seg_azimuths`: I always get the a residual close to zero between pyproj and matlab? I think the preprocessing in `celeri.build_model` fixes the differences? We could compare with the raw data by loading with `celeri.read_data`, and compare with that if we want.
- [ ] `doc_tri_coupling`: I get a singular matrix error in the pre-refactor state.
- [ ] `doc_matplotlib_vis`: This had different results in the commited plots compared to what I got after rerunning the cells in the pre-refactor state.
- [ ] `demo_classic_tikhonov`: I get the same results pre- and post-refactor, but slightly different compared to last commited results.
- [ ] `demo_qp_dwe_coupling_bound`
- [ ] `demo_qp_dwe_slip_bounds`
- [ ] `demo_classic_tikhonov_slip_bounds`
- [ ] `doc_dense_vs_hmatrix`
- [ ] `doc_hmatrix_tests`
- [ ] `doc_hmatrix`

brendanjmeade:
Gigantic indeed!

gigantic

PR #150, opened 2025-04-30
Euler_initial

brendanjmeade:
Initial implmentation of Euler pole and uncertainty calculation.

These are pretty direct translations from the old blocks code that @jploveless pointed me to.

The uncertainty calcuation is there but I pass it a covariance matrix full of zeros because I'm not creating a covariance any more with a QP solve.  I should pass something for the classical solve case.

I'm pretty sure the conversion of rate to degrees per million years is wrong.  What do you think @jploveless?

### DIFF
--- a/celeri/celeri.py
+++ b/celeri/celeri.py
@@ -4466,7 +4466,7 @@ def rotation_vectors_to_euler_poles(
     euler_lon = np.where(euler_lon < 0, euler_lon + 360, euler_lon)
 
     # Convert the rotation rate from rad/yr to degrees per million years
-    SCALE_TO_DEG_PER_MILLION_YEARS = 1e6  # TODO: Check this
+    SCALE_TO_DEG_PER_MILLION_YEARS = 1e3  # TODO: Check this
     euler_rate = SCALE_TO_DEG_PER_MILLION_YEARS * np.rad2deg(euler_rate)
 
     return euler_lon, euler_lat, euler_rate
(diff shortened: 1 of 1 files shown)


jploveless:
I looked into this a while ago, I think when trying to apply a priori block rotation constraints for Juan de Fuca. The [note](https://github.com/brendanjmeade/celeri/blob/30c84a86f1abe757844d420a87e1c654a97d2368/celeri/celeri.py#L61) in the constants section of `celeri.py` says

`DEG_PER_MYR_TO_RAD_PER_YR = 1 / 1e3`
The conversion should be 1 / 1e3. Linear units for Cartesian conversions are in meters, but we need to convert them to mm to be consistent with mm/yr geodetic constraints units. Rotation constraints are expressed in deg\/Myr, and when applied in celeri we are effectively using m\*rad\/Myr. To convert to the right rate units, we need 1e-3\*m\*rad\/Myr. This conversion is applied in `get_data_vector` (JPL 12/31/23)

When calculating Euler pole rotation rates, we need to take the reciprocal because we're going from rad to deg, so I think  line 4469 of the PR should be changed to:  

`SCALE_TO_DEG_PER_MILLION_YEARS = 1e3`

(I was going to comment on this in the review request, but I've never done that inline, so I'm playing it safe here!)

brendanjmeade (commented):
Thanks @jploveless!  Corrected in the latest commit.  Progress!

jploveless:
While I think this is correct, I'll try to give it a check in the coming days. I figure a good test is that the Cascadia estimate should yield a Juan de Fuca rotation rate that's similar to the input!

brendanjmeade:
That's a great idea.  I'll merge for now and I'm happy to iterate.  I really appreciate the pointers and the review.

PR #152, opened 2025-05-02
Added propagation of rotation vector uncertainties to Euler pole uncertainties for classical case.

brendanjmeade:
Add linearized uncertainy rotation vector to Euler pole uncertainy propagation. Classical case with spatial smoothing only.  I haven't done this for the the eigenmode case because we don't reliably have a covariance matrix for that case.  I could check for this and propagate if one exists.


jploveless:
`model_block.csv` from a WNAM run writes a Juan de Fuca pole that agrees in terms of coordinates and uncertainties with the a priori pole. The coordinates are written as the antipode, but with the same sign of rotation as the input, so I think we have a sign flip somewhere in the code. 

Input: 244.66±0.1, 33.92±0.1, -1.513±0.001
Output: 64.6600±0.1000, -33.9200±0.1000, -1.5130±0.0010

`Blocks` seems to also write the antipode but correctly flips the sign of the rotation rate to positive. I'm actually not sure how the written negative rate is possible given the calculation as just a magnitude of the Euler vector components.

jploveless:
All other rotation rates in `model_block.csv` are positive; only the JDF is negative. This suggests to me that somehow the a priori rotation rate is getting inserted into the `estimation` dict that is written as output, but I can't see where, and it's funny that the original longitude and latitude of the a priori pole are _not_ retained (because the antipode is written).

brendanjmeade:
> All other rotation rates in `model_block.csv` are positive; only the JDF is negative. This suggests to me that somehow the a priori rotation rate is getting inserted into the `estimation` dict that is written as output, but I can't see where, and it's funny that the original longitude and latitude of the a priori pole are _not_ retained (because the antipode is written).

Thanks for looking into this @jploveless!

I think I have an explanation!  There are different columns with rotation rates in them.  I don't know what is supposed to be in "rotation_rate" but I definitely see the  -1.5130 rate there.  However, the Euler pole rotation rates that I intentionally calculate are in the column with the header "euler_rate" and there, the JdF rate is positive, 1.5130, as are all the rates in that column.

In summary:
- The column named "euler_rate" is the one that I added in this PR, and it contains intentionally calculated Euler pole rotation rates.  
- I don't know what's supposed to be in the "rotation_rate" column.
- This means that we are recovering the correct sign for the JdF prior!
- This leads me to assert that there isn't a sign correction to be made and that the Euler pole locations and rates are being calculated correctly, albeit with absurdly small errors.

Thoughts?

jploveless:
Thank you for the explanation! Now I'm seeing what's going on, I think, and it's great news that the sign of the output rotation for JDF is correct, given the antipode coordinates. 

What do you think about trying to consolidate the fields that are written to `model_block.csv`? There are already fields for Euler pole coordinates and rates, as well as uncertainties on all of them, though we have some near-duplication of `euler_lon_sig` and the new `euler_lon_err`, etc. We could just overwrite all Euler pole fields in the `block` dictionary with the calculated values, then write that to `model_block.csv`. This would mean there's no output record of any _a priori_ poles, but I feel that that would be okay, given that the file is meant to represent estimated model parameters. I think this would be as simple as a renaming of calculated variables in the Euler pole location/rate and uncertainty propagation functions, making them match the fields that were read from the input `*.block` file.

brendanjmeade:
@jploveless 

> Thank you for the explanation! Now I'm seeing what's going on, I think, and it's great news that the sign of the output rotation for JDF is correct, given the antipode coordinates.

Agreed.

> What do you think about trying to consolidate the fields that are written to `model_block.csv`? There are already fields for Euler pole coordinates and rates, as well as uncertainties on all of them, though we have some near-duplication of `euler_lon_sig` and the new `euler_lon_err`, etc. We could just overwrite all Euler pole fields in the `block` dictionary with the calculated values, then write that to `model_block.csv`. This would mean there's no output record of any _a priori_ poles, but I feel that that would be okay, given that the file is meant to represent estimated model parameters. I think this would be as simple as a renaming of calculated variables in the Euler pole location/rate and uncertainty propagation functions, making them match the fields that were read from the input `*.block` file.

I want to keep everything we have.  Goodness knows, we may want it in the future.

What I've done with the latest commit is add "_clp" shorthand for "classical linear propagation" to all of the Euler pole predictions.

`euler_lon` -> `euler_lon_clp`
`euler_lat` -> `euler_lat_clp`
`euler_rate` -> `euler_rate_clp`
`euler_lon_err` -> `euler_lon_err_clp`
`euler_lat_err` -> `euler_lat_err_clp`
`euler_rate_err` -> `euler_rate_err_clp`

This is intended to indicate that these fields are distinct.  I think this solves the nomenclature issue that you brought up.  Also, I'm using "err" for these rather than "sig" because I'm planning to add some non-linear error/uncertaitny estimates.

I think we're good to merge.

jploveless:
I like that separation! It makes it clear what might is specified as an _a priori_ constraint vs. what's estimated.

Issue #154, opened 2025-05-08
Experiments with constraint relaxation


maresb:
I was worried that our approach with constraints was preventing us from achieving the global optimum. Apparently not. 🚀 

```python
    annealing_schedule = [32, 16, 8, 4, 2, 1, 0.5, 0.25, 0.125]  # mm/yr

    if num_oob > 0:
        looseness = 0
    else:
        if len(annealing_schedule) > 0:
            looseness = annealing_schedule.pop(0)
            print(f"Loosening constraints by {looseness}")
        else:
            raise StopIteration
```

And then when setting the bounds:
```python
            target = kinematic
            diff = upper - target
            upper = target + factor * diff + looseness
            diff = lower - target
            lower = target + factor * diff - looseness
```

Here are the iterations with no out-of-bounds:
```python
[(12, np.float64(42.74860926501867)),
 (24, np.float64(42.753248031645)),
 (35, np.float64(42.747924398620576)),
 (45, np.float64(42.74775954803366)),
 (53, np.float64(42.74062899173406)),
 (64, np.float64(42.73733040276304)),
 (72, np.float64(42.737317450844806)),
 (80, np.float64(42.737317450844806)),
 (87, np.float64(42.737317450844806)),
 (93, np.float64(42.737317450844806))]
```

And here's a video of the experiment, with the zero-contour shown:

https://github.com/user-attachments/assets/728c28b0-939b-494a-9420-724da28091af

maresb:
It's really robust to changing the annealing, always converging to the optimum at `42.737317450844806`:

```python
annealing_schedule = [0.25, 0.125, 0.0625]

[(12, np.float64(42.74860926501867)),
 (21, np.float64(42.739516690393124)),
 (31, np.float64(42.73733186436606)),
 (35, np.float64(42.737317450844806))]
```

aseyboldt:
So without the annealing it was still a little bit worse than after annealing?
Maybe we should enable a few annealing steps by default? They don't cost that much after all.

maresb:
It does cost though. Remembering the 0-indexing, it goes from 13 iterations to 36 iterations, so almost 3x as long.

maresb:
Another run:
```python

annealing_schedule = [0.125, 0.125, 0.125]

[(12, np.float64(42.74860926501867)),
 (21, np.float64(42.73955301805384)),
 (31, np.float64(42.73733186733481)),
 (36, np.float64(42.737317450844806))]
```

[...]

PR #157, opened 2025-05-15
Add support for annealing

maresb:
Adds an `annealing_schedule` argument to `celeri.optimize.minimize()`.

Specify a list of looseness values (mm/yr) for relaxing the constraints after the out-of-bounds values have all been fixed. Example, `annealing_schedule=[0.125, 0.125, 0.125]` will result in the following `qr_sum_of_squares` losses, filtered to show only those iterations that have settled to no out-of-bounds:

(0-indexed iteration number, residual 2-norm):
```python
[(12, np.float64(42.74860926501867)),
 (21, np.float64(42.73955301805384)),
 (31, np.float64(42.73733186733481)),
 (36, np.float64(42.737317450844806))]
```

For more details and visualizations, see #154.

### DIFF
--- a/celeri/optimize.py
+++ b/celeri/optimize.py
@@ -1049,7 +1049,7 @@ class MinimizerTrace:
         print(f"Iteration: {iter_num}")
         print(f"{oob} of {total} velocities are out-of-bounds")
         print(f"Non-convex constraint loss: {nonconvex_loss:.2e}")
-        print(f"residual 2-norm: {objective:.5e}")
+        print(f"residual 2-norm: {objective}")
         print(f"Iteration took {iter_time:.2f}s")
         print()
 
--- a/celeri/optimize.py
+++ b/celeri/optimize.py
@@ -944,11 +944,19 @@ def _tighten_kinematic_bounds(
     factor: float = 0.5,
     iteration_number: int,
     num_oob: int,
+    annealing_schedule: list[float],
 ):
     assert factor > 0 and factor < 1
 
-    if num_oob == 0:
+    if num_oob > 0:
+        looseness = 0
+    elif len(annealing_schedule) == 0:
         raise MinimizationComplete()
+    else:
+        # We have fixed all OOBs, and annealing is enabled, so use the first remaining
+        # value in the annealing schedule as the loosenenss.
+        looseness = annealing_schedule.pop(0)
+        print(f"Loosening constraints by {looseness}")
 
     def tighten_item(limits: VelocityLimitItem, coupling: CouplingItem):
         estimated = coupling.estimated_numpy()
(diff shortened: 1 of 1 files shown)


maresb:
@aseyboldt, this is effectively a refactor that adds an extra argument to `minimize`. The commits are structured, so I hope this is quick and easy to merge.

My plotting and video functionality is in a very messy notebook in [a separate branch](https://github.com/maresb/celeri/tree/annealing-plots) that I should clean up at some point, but that's for a separate PR.

aseyboldt:
This looks good to me, but maybe we should provide a default annealing schedule? We could for instance have two arguments, `annealing: bool` and `annealing_schedule: list[int] = [something]`

aseyboldt:
The problem I see with that is that unless we turn it on by default, every time someone wants to enable annealing, they would have to decide what schedule is sensible. I think it is important that there is a pre-configured schedule. Nobody will have a better guess what makes sense than you. ;-)

brendanjmeade:
Thanks for the ideas @maresb and @aseyboldt.  Some notes:

- I don't think the annealing approach will be used in practice.


- I agree with @aseyboldt's proposal that there should be separate logic branches for the annealing and non-annealing cases.  I do not want to do the case where the no annealing case is just a schedule of `[]`.  Annealing could be completely removed at a future date, and I want this to be easy.  Also, as mentioned above, the annealing may not be used much in practice, and I don't want users to have to understand why the QP solve is wrapped inside of an annealing process (even an empty one).


- I don't want this to become a research problem, so if there's no sensible default schedule and we've already learned that for the cases we've considered, we're close to a global minimum, we should consider removing the annealing.

maresb:
Ok, I will look when I get the chance to see if I make some separation between the main logic and annealing logic to make it easy to remove.

brendanjmeade:
Great @maresb.  The vanilla QP logic should be cleanly implemented as option 1.  I think this is important as @aseyboldt is soon to be revisiting this to make the QP estimation work without the doubling of constraints.

**A clean implementation of the iterative QP with the proper number of constraints is a very high-priority goal.**

brendanjmeade:
Are passing on this for the sake of wrapping up the project?  Lines have to be drawn!

maresb:
Indeed, this is my inclination. I think there's potentially a very interesting work packet here, but I think it's inessential to the immediate objectives. Also, for our current solution, we discovered that the difference this makes is infinitesimal.

Let's discuss later how this fits into broader objectives, and we can decide to what extent we revive this effort.

> Lines have to be drawn!

Yes, especially for making beautiful contour plots! And if you want accurate zero-contours for strike-slip then this is actually fairly critical.

PR #162, opened 2025-06-03
Setting burial_depth = 0 in all segment files

brendanjmeade:
Also burial_depth_sig and burial_depth_flag


brendanjmeade:
I have no idea why any of the `burial_depth`s were non-zero.  This PR sets them to zero in all `*segment.csv` files now.

@maresb They should be assumed to be zero (or better yet nonexistent!) inside the celeri codebase.  It's simply a legacy concept that we don't need.

@maresb @aseyboldt This will likely change results from the baseline that we had before.

brendanjmeade:
> Seems reasonable to me. I wonder if just adding `segment["burial_depth"] = np.zeros(len(segment))` to `celeri.process_segment` could be a way of addressing this going forward

@jploveless I think that's reasonable, but I think @maresb is likely to do one better and simply remove all references to it in the elastic calculation code.

> (suggested in part because my quick attempt to remove all references to burial depth in celeri_ui failed!)?

@jploveless I'm interested to hear more about this.  Did you just try to remove all instances of "burial_depth" from the celeri_ui repo?  I recall that segment file column names are spread across several files in celeri_ui.  I have successfully added column names to celeri_ui.

jploveless:
I think it's that local changes to `celeri_ui` don't affect the GitHub.io hosted version. Issue created [here](https://github.com/brendanjmeade/celeri_ui/issues/45).

PR #167, opened 2025-06-09
Organizing and cleaning input files

brendanjmeade:
This PR does three things focused on organizing and cleaning input files necessary to run the `demo*.ipynb` notebooks:

1. Eliminates all input files other than those necessary for the `demo*.ipynb` notebooks.  These fit into two categories:
   - Western North America.  All files other than mesh files with geographically specific names are prefixed with `wna`.
   - Japan.  All files other than mesh files with geographically specific names are prefixed with `japan`.
2. Eliminates extraneous fields in `*mesh.json` files for following suggestions from @jploveless's [wiki page](https://github.com/brendanjmeade/celeri_notes/wiki/Mesh-parameter-descriptions).  Notes:
   - The `coupling_constraint*` fields are still there for now, but should be removed later.
   - `*mesh.json` naming replaces `*mesh_params.json` and `*mesh_parameters.json`.
3. `config` replaces `command`. This is part of a larger standardization led by @aseyboldt.  Notes:
   - This is done to variable names in the `demo*.ipynb` notebooks.
   - `*config.json` replaces `*command.json` files.
   - `celeri.get_command()` is still used in the `demo*.ipynb` notebooks but will be replaced with `celeri.get_config()` once the refactoring with data classes is ready.


jploveless:
In [`*mesh.json`](data/mesh/wna_mesh.json), should the `qp_mesh_tde_bound` and `mesh_tde_coupling_bound` flags remain to indicate which, if either, constraint should be applied? At present I think these are only used in the demo notebooks, but in an eventual `celeri_solve` it'd be nice to be able to have some control over the constraint style from the `*mesh.json` (and I'm not sure how things would respond if there were non-zero upper, lower bounds for both slip rates and coupling).

aseyboldt:
Using "0" wouldn't be an improvement. The idea would be to use something
like global_elastic_cutoff_method = "quantile" or whatever might make sense.
Or for a mesh, it might be `mesh_kind = "subduction_zone"`. In the
dataclass you can then specify `mesh_kind: Literal["subduction_zone",
"mantel_bounday", "segment_mesh"]`, to make clear what the allowed options
are.
I think all integer flags would ideally be either a bool, or something like
that.
Integer flags are a disaster waiting to happen. You never know what the
options are, you always have to double check what the values mean and it is
hard to keep code up to date if the valid options change at some point (ie
some old code only handles values 0 and 1, but gets handed a 2...)
I realise some external code in the ui for instance might depend on the
current config, and that we might not want to change that now though.

brendanjmeade:
I hear you and appreciate your points @aseyboldt.  

> Using "0" wouldn't be an improvement. The idea would be to use something like global_elastic_cutoff_method = "quantile" or whatever might make sense. Or for a mesh, it might be `mesh_kind = "subduction_zone"`. In the dataclass you can then specify `mesh_kind: Literal["subduction_zone", "mantel_bounday", "segment_mesh"]`, to make clear what the allowed options are. I think all integer flags would ideally be either a bool, or something like that. Integer flags are a disaster waiting to happen. You never know what the options are, you always have to double check what the values mean and it is hard to keep code up to date if the valid options change at some point (ie some old code only handles values 0 and 1, but gets handed a 2...).  I realise some external code in the ui for instance might depend on the current config, and that we might not want to change that now though.

I certainly think what you've described is much more readable and self-documenting.

This type of change would touch not only celeri but parsli, result_manager, and celeri_ui as well as some ancient scripts/notebooks that aren't a core part of celeri.  Some things that are currently integer flags are treated as "booleans" and other things that we treat as flags mean specific things for specific integer values.

I'm just not ready to take on this change across the codebase at this time and feel that there are more pressing goals.  We can certainly discuss this again after we've ticked off a few more goals.

aseyboldt:
Looks good to me now :-)

PR #179, opened 2025-06-16
fix: don't double constraints in optimizer

aseyboldt:
This removes the duplicate constraints in the new optimizer. On my (slow) machine, the japan example now runs in 1min instead of 3min.

### DIFF
--- a/celeri/operators.py
+++ b/celeri/operators.py
@@ -1824,37 +1824,42 @@ def get_qp_tde_inequality_operator_and_data_vector(
 
 
 def get_qp_all_inequality_operator_and_data_vector(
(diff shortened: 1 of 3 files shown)


brendanjmeade:
The solve runs for me in 1 minute 41 seconds on an M4 Macbook.

 
Screenshot 2025-06-16 at 3 57 54 PM

aseyboldt:
I'll fix the tests tomorrow :-)

brendanjmeade:
Sounds good!  I think I may have caused some issue pushing (or not pushing a pixi.lock file)

There's so much good stuff happening right now!

PR #188, opened 2025-06-17
chore: update pixi lock file

aseyboldt:
fix the tests


aseyboldt:
I think something was broken in the lock file, not sure what to be honest. But I thought updating is a good idea anyway, and that fixed the issue.

brendanjmeade:
I may have messed something up.  I keep getting a changed `pixi.lock` file locally after some edits.  I don't understand why.  Guidance welcome!

PR #190, opened 2025-06-20
Implement serialization

aseyboldt:
This implements serialization without pickle.

Each of the dataclasses gets a `to_disk` and `from_disk` method, that writes the content to a combination of json and zarr.

A helper function `output.dataclass_to/from_disk` does most of the work, and handles all ndarray or json-serializable fields. Other fields (like other dataclasses) need to be handeled manually in the corresponding `to_disk` method.

I also converted the config dataclasses (so the model config and the mesh configs) to use pydantic, which makes the json serialization of those easier.

Most of the `write_output` functionality is now in `Estimation`. You can access the modified dataframes like `station` with the model estimation as `estimation.station`.

A written run now looks like this:

```
/
estimation = trace_japan.minimizer.to_estimation()
!tree {estimation.model.config.output_path} -L3

../runs/0000000072
├── arrays.zarr
│   ├── data_vector
│   │   ├── c
│   │   └── zarr.json
│   ├── operator
│   │   ├── c
│   │   └── zarr.json
│   ├── state_vector
│   │   ├── c
│   │   └── zarr.json
│   ├── weighting_vector
│   │   ├── c
│   │   └── zarr.json
│   └── zarr.json
├── model_block.csv
├── model_meshes.csv
├── model_mogi.csv
├── model_segment.csv
├── model_station.csv
├── operators
│   ├── arrays.zarr
│   │   ├── block_motion_constraints
│   │   ├── block_strain_rate_to_velocities
│   │   ├── global_float_block_rotation
│   │   ├── mogi_to_velocities
│   │   ├── rotation_to_slip_rate
│   │   ├── rotation_to_slip_rate_to_okada_to_velocities
│   │   ├── rotation_to_tri_slip_rate
│   │   ├── rotation_to_velocities
│   │   ├── slip_rate_constraints
│   │   ├── slip_rate_to_okada_to_velocities
│   │   └── zarr.json
│   ├── eigen
│   │   ├── arrays.zarr
│   │   └── remaining.json
│   ├── index
│   │   ├── arrays.zarr
│   │   ├── eigen
│   │   ├── remaining.json
│   │   └── tde
│   ├── model
│   │   ├── block.parquet
│   │   ├── config.json
│   │   ├── meshes
│   │   ├── mogi.parquet
│   │   ├── sar.parquet
│   │   ├── segment.parquet
│   │   └── station.parquet
│   ├── remaining.json
│   ├── smoothing_matrix_0.npz
│   ├── smoothing_matrix_1.npz
[...]

PR #194, opened 2025-06-23
test: add serialization test for Estimation

aseyboldt:
Followup from #190, to add some missing tests.

There's also something I should mention about the new format:

With the `to_disk` / `from_disk` methods we get a data format that I think we can be pretty sure people will be able to read relatively easily for years to come, *if they read the arrays and json manually using the json or zarr modules*. The `from_disk` method will *not* work long-term on old datasets, unless we take care of this every time we add non-optional attributes to a dataclass.

So let's say we add a non-optional new field to `model`. Then the `Model.from_disk` method will fail to read an old run that didn't have that attribute yet. (I think that's mostly a good thing, because if we want to use that model with new code, it would be strangely incompatible with the new code). We can do some extra work to ensure that we *can* read old objects, by adding a manual check to `from_disk` every time that we add an attribute, and handle the case of that missing attribute in `from_disk` on that dataclass.

So after we fix the config up a little more, I think it would be a good idea to add a small run to the tests folder, and make a test that tries to read this with `from_disk`. That test will then fail when we break deserialization of old runs, and tell us what we have to do to repair it.

We could also additionally store a pickle version of the `Estimation`. Reading that might break in the future at some point, and we can never reliably use an unpickled Model in new code, to for instance run some new optimization algorithm on it (because of the same issue of changing attributes). But if you just want to get your hand on the old model and run a few print statements, that might be useful from time to time.


maresb:
If we want to get fancy we could add a schema version integer. That way, whenever we make a backwards-incompatible change, we raise an error telling the user to upgrade celeri.

If we want to get super-fancy, we can support previous schemas and define migrations.

But I want to also warn that we should be vigilant against scope creep. So the fancy stuff should probably wait for a SOW2. Besides, I'm eager to start making progress with the uncertainty estimation.

PR #196, opened 2025-06-23
Repair celeri_solve and some cleanup

aseyboldt:
Clean the config a bit, based on the wiki info.

Change the names of runs from integers like `00000000005` to time stamps like `2025-06-23T12-25-01Z`. This avoids race conditions during notebook execution.

Some cleanup for the elastic operator caching.

The commits are pretty independent, probably easiest to read them one commit at a time.

### DIFF
--- a/celeri/config.py
+++ b/celeri/config.py
@@ -123,39 +123,36 @@ class Config(BaseModel):
 
 
 def _get_output_path(base: Path) -> Path:
-    """Generate a unique timestamped output path within the base directory.
+    """Generate a unique numbered output path within the base directory.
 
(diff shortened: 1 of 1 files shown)


aseyboldt:
I was running into race conditions when I started several notebooks at the same time with the new script.
But if you prefer the numbers, I think we can also avoid the race conditions there.

brendanjmeade:
Let's go for ordered numbers if that's feasible.  So easy to remember critical runs.  Do you want me to merge this PR now or wait for that?

brendanjmeade:
Sounds like a great plan.

PR #197, opened 2025-06-24
Cleaner and more expressive constraint config

aseyboldt:
@brendanjmeade I tried implementing the constraint handling we discussed last meeting. Does this capture the constraints that we want to specify?

The new config is in `MeshConfig`, together with some comments. Hopefully, the meaning of the parameters is clear from those comments, if it isn't we can at least use that as an exercise to improve them :-)

Does this new config make sense to you? Are the names ok or is there something we should change about those?

Along the way I also updated the nomenclature in `optimize.py` a bit, and changed the path handling in the config files, such that the paths to other files are relative to the config file where they are used, instead of relative to whatever happens to be the current working directory. (I had a couple of `run` and `operator` folders all over the place because of that ambiguity...)


brendanjmeade:
@aseyboldt It seems to!

PR #198, opened 2025-06-24
Segmesh dev

jploveless:
I had to make a few changes to `segmesh.py` to agree with current syntax


jploveless:
@aseyboldt I'm trying to write a new `*mesh_params.json`, updated with new entries for newly created meshes. I think I'm updating the list of `MeshConfig` objects in `model.config.mesh_params` correctly, but I'm struggling to write to disk. I've tried three approaches, starting in [line 285 of `segmesh.py`](https://github.com/brendanjmeade/celeri/blob/ae210f15daf1a07631dc5060cbf3293d3f10bbbf/celeri/scripts/segmesh.py#L285), and all of them give me different errors that I describe in the commented lines. pydantic is brand new to me, so I would appreciate some guidance. 

https://github.com/brendanjmeade/celeri/pull/198/commits/ae210f15daf1a07631dc5060cbf3293d3f10bbbf includes two new files that I'm using as a test of the script, which I run from `celeri/celeri` using: 

```
python ./scripts/segmesh.py ../data/config/japan_config_test.json
```

aseyboldt:
You can first dump the individual configs to a json-compatible dict, put them in a list and then serialize that list using json:

```python
import json

data = [mesh_config.model_dump(mode="json") for mesh_config in config.mesh_params]
json.dumps(data, indent=4)
```
The `mode="json"` makes sure that pydantic handles for instance `pathlib.Path` objects, which you normally can't directly serialize with the json module.

If #197 goes through, the paths in the config file will be interpreted as relative to the location of the config file. The above would write absolute paths into the mesh config file, so you wouldn't be able to move it somewhere else. So maybe it would be nicer to transform those paths to relative paths. Something along the lines of

```python
for mesh_config in config.mesh_params:
    new_file_location = pathlib.Path(new_file_location).resolve()
    for name in mesh_config.model_fields:
        var = getattr(mesh_config, name)
        if isinstance(var, pathlib.Path):
            setattr(mesh_config, name, var.relative_to(new_file_location))
```

PR #208, opened 2025-07-11
removing used fields from def make_default_segment()

brendanjmeade:


brendanjmeade:
@aseyboldt @jploveless 

This is a small set of edits to remove unused fields from the `make_default_segment` function, done as a part of some general segment-related cleanup.

I've noted that `make_default_segment` is defined twice in the codebase.

1. In `model.py`
2. In `snap_segments.py`

Before merging, I wanted to ask if we need to define this twice.  Can we import it into `snap_segments.py` so that it doesn't have to be redefined?  I know that you two had some discussion that I think was related to scripts like `snap_segments.py` (?), so I thought I'd check in and get your opinions.

brendanjmeade:
Ping @aseyboldt @jploveless

PR #209, opened 2025-07-14
Adapt doc_matplotlib_vis_widget notebook to refactor

aseyboldt:
@brendanjmeade Hope that helps with reading in the results from previous runs?


brendanjmeade:
Totally works!  I'm currently working on PR that does a wide range of standard visualizations.

Issue #186, opened 2025-06-17
Incorrect constraint fix in #179


aseyboldt:
I've been going through the change in https://github.com/brendanjmeade/celeri/pull/179 again, and I think I didn't fix it correctly yet. The current code does not properly use the mesh-specific constraints on the slip rates.

@brendanjmeade just wanted to let you know I'm still working on this.

aseyboldt:
After our discussion yesterday, I think the current state is not broken, but we might want to add the ability to add scientific constraints on the elastic slip rates. (in contrast to the constraints that were previously added on the estimation purely during the sqp algorithm)

brendanjmeade:
@aseyboldt Following up on the issue of coupling on a vertical fault, I'm writing to share an example notebook and input files.  These can be found in:

https://github.com/brendanjmeade/anatolia_block_models_coupling

This example has a vertical fault where I kind of hand-tune the notebook to only consider the coupling values for the strike-slip case.  Effectively, for the sequential quadratic programming part, I only consider the strike-slip components for the number of out-of-bounds elements, `n_oob`

```python
    # HACK: Consider strike slip case only because coupling isn't well defined fro DS on a vertical fault
    n_oob = len(tde_coupling_ss_lower_oob_idx) + len(tde_coupling_ss_upper_oob_idx)
```

This does the job, but is super hacky.

I think the easiest way to explain would be via ~15 minute Zoom if you have the time.

aseyboldt:
How about a config like this? Would this be able to capture everything we need?

```python
class CouplingConstraint(BaseModel):
    lower: float | None = None
    upper: float | None = None


class ElasticConstraint(BaseModel):
    lower: float | None = None
    upper: float | None = None
    # Override the lower and upper bounds for specific indices
    lower_index_override: list[int] = []
    lower_override_values: list[float] = []
    upper_index_override: list[int] = []
    upper_override_values: list[float] = []


class MeshConfig(BaseModel):
    ...
    # Constraints for the coupling on the mesh, ie the ratio
    # of elastic and kinematic slip rates.
    # Specified independently for strike-slip and dip-slip
    coupling_constraints_ss: CouplingConstraint = CouplingConstraint(
        lower=None, upper=None
    )
    coupling_constraints_ds: CouplingConstraint = CouplingConstraint(
        lower=None, upper=None
    )

    # Constraints for elastic rates on the mesh.
    elastic_constraints_ss: ElasticConstraint = ElasticConstraint(
[...]

PR #223, opened 2025-07-22
Update `segmesh.py` for single flag; make change to `operators.py`

jploveless:
`segmesh.py` now looks for a single flag (`create_ribbon_mesh = 1`) for segment meshing, rather than unique integers set in `celeri_ui`. Distinct ribbons are identified based on unique pairs of block labels. The slight downside is that ribbons cannot smoothly span triple junctions, but I feel that that’s more consistent with the coupling constraints, because the kinematic rates on any ribbon are clearly associated with relative motion of the bounded blocks. 

[The change to `operators.py`](https://github.com/brendanjmeade/celeri/blob/6d3f062d152bb15fb277c9df5e29c3806ed7c711/celeri/operators.py#L866) is in how the coupling constraints are applied to meshes. Previously, `_store_tde_coupling_constraints` looped as 

```
for mesh_idx in range(np.max(model.segment.mesh_file_index) + 1):
```

but this presented a problem when, for example, the CMI is `model.meshes[3]` but `segmesh.py` added `model.meshes[4]`. In that case, the loop above looks for `model.segment.mesh_file_index = 3` but that doesn't exist because the CMI isn't connected to segments. `operators.py` now loops as

```
for mesh_idx in np.unique(model.segment.mesh_file_index[model.segment.mesh_flag == 1]):
```

which should only loop through segment-associated meshes.


brendanjmeade:
@jploveless 

> `segmesh.py` now looks for a single flag (`create_ribbon_mesh = 1`) for segment meshing, rather than unique integers set in `celeri_ui`. Distinct ribbons are identified based on unique pairs of block labels. The slight downside is that ribbons cannot smoothly span triple junctions, but I feel that that’s more consistent with the coupling constraints, because the kinematic rates on any ribbon are clearly associated with relative motion of the bounded blocks.

You've already made the case during out last Zoom chat.  You were right then and you are right now.

> ```
> for mesh_idx in np.unique(model.segment.mesh_file_index[model.segment.mesh_flag == 1]):
> ```
> 
> which should only loop through segment-associated meshes.

This is very thoughtful and makes good sense.  We could try some more explicit flagging but this is nice and dynamic so I vote we stay with it.

### Note regarding .msh files.
-  Currently, I don't think all of these new `.msh` files should be added to the repo.
-  Keeping them would make sense if they are part of a test.
-  I think they are better a part of the japan_imaging repo
-  Happy to consider this!

jploveless:
> You've already made the case during out last Zoom chat. You were right then and you are right now.
 
I wanted to document my thinking somewhere, so I just pasted this in from my text to you, but I realize a PR comment is not necessarily the best place to permanently document ideas! 

> * Currently, I don't think all of these new `.msh` files should be added to the repo.

I'm sorry, I meant to strip these out before submitting the PR. They're removed in the latest commit. But that commit is still failing in the serialize operators test, so I'll do some more troubleshooting tomorrow.

jploveless:
It seems like there are a few places in `solve_sqp` where meshes are looped through using a `range(n_segment_meshes)`, where `n_segment_meshes = np.max(model.segment.mesh_file_index).astype(int) + 1` rather than explicitly identifying segment-replacing meshes that would be eligible for coupling constraints. I am very reluctant to modify `solve_sqp` given all of the work you all have put into it, so maybe a better alternative is to just have a policy of listing CMI meshes last in the `*_mesh.json` so that segment-replacing meshes are a sequential list at the top of the parameter file and therefore correspond to the `range(n_segment_meshes)`. I can rewrite `segmesh.py` to do that. 

A downside to this is that `mesh_flag` can't be fully used, since we couldn't run a model where we set `mesh_flag = 0` without some more explicit definitions like I wrote above in `operators.py`.

brendanjmeade:
> It seems like there are a few places in `solve_sqp` where meshes are looped through using a `range(n_segment_meshes)`, where `n_segment_meshes = np.max(model.segment.mesh_file_index).astype(int) + 1` rather than explicitly identifying segment-replacing meshes that would be eligible for coupling constraints. I am very reluctant to modify `solve_sqp` given all of the work you all have put into it, so maybe a better alternative is to just have a policy of listing CMI meshes last in the `*_mesh.json` so that segment-replacing meshes are a sequential list at the top of the parameter file and therefore correspond to the `range(n_segment_meshes)`. I can rewrite `segmesh.py` to do that.

Thanks for the discussion @jploveless.  Let me try to state a few things to make sure we're on the same page:

0. In the original @brendanjmeade version, we had clearly assumed that coupling bounded meshes were the first `n_segment_meshes` and that meshes listed after that in a `*_mesh.json` file were only slip rate constrained.
1. `n_segment_meshes` gives the number of non-BEL/CMI meshes
2. All meshes have either slip rate or coupling constraints.
3. The type of constraint is specified in a `*_mesh.json` file with lines like
```json
        "elastic_constraints_ss": [-50, 50],
        "elastic_constraints_ds": [-50, 50],
        "coupling_constraints_ss": [0, 1],
        "coupling_constraints_ds": [0, 1],
```
4. The logic for dealing with the selection of either slip rate or coupling constraints was discussed in https://github.com/brendanjmeade/celeri/issues/186

Comments:
- As far as I can tell, I don't believe we've brought the JSON into consistency with what was discussed in the issue.  Specifically, I think we should be setting:
```json
        "coupling_constraints_ss": [null, null],
        "coupling_constraints_ds": [null, null],
```
for any mesh with slip rate constraints but not coupling constraints.  If that logic is present, we could simply loop over all meshes, regardless of their order, because the decision about slip rate or coupling bounds would not depend on the ordering of the meshes.
- In the context of the above, I think there is still some ambiguity to be resolved about how to loop over which meshes and specify coupling vs slip rate constraints.


> A downside to this is that `mesh_flag` can't be fully used, since we couldn't run a model where we set `mesh_flag = 0` without some more explicit definitions like I wrote above in `operators.py`.

I think I agree.  Is there a case where we'd want to run with `mesh_flag = 0`?  I think that's required to not do a rectangular elastic calcuation (which then gets replaced by a mesh).

jploveless:
> Is there a case where we'd want to run with `mesh_flag = 0`? I think that's required to not do a rectangular elastic calcuation (which then gets replaced by a mesh).

I think it would be nice to be able to toggle meshes off, but not necessary. I'm testing some runs right now where I want to incrementally add segmeshes, and the workflow for doing so is to edit segment properties in `celeri_ui`, run `segmesh.py`, manually edit `*_mesh_param_segmesh.json` to apply coupling constraints, then run the sqp notebook. Having `mesh_flag` as a functional flag would eliminate a few of those steps.

brendanjmeade:
> I think it would be nice to be able to toggle meshes off, but not necessary. I'm testing some runs right now where I want to incrementally add segmeshes, and the workflow for doing so is to edit segment properties in `celeri_ui`, run `segmesh.py`, manually edit `*_mesh_param_segmesh.json` to apply coupling constraints, then run the sqp notebook. Having `mesh_flag` as a functional flag would eliminate a few of those steps.

I'm working to understand this.  I view the primary role of `mesh_flag` as an indicator to "turn off" particular rectangular dislocation elements on a segment-by-segment basis.

Am I right in thinking that you are looking for something like "Don't use a particular mesh?"

brendanjmeade:
@jploveless 

> Your statements here make me realize that I've always thought about `mesh_flag` incorrectly! I indeed have used it in `Blocks` to turn off a particular mesh, by setting `mesh_flag = 0` for **all** segments that have a particular `mesh_file_index` value. This is clunky and ripe for misuse, because it makes it possible for there to be essentially doubled elastic contribution. For example, if not all segments along a mesh have `mesh_flag = 0`, they would have an elastic contribution from their rectangle geometry as well as that section of the mesh, because the logic of `zero_mesh_segment_locking_depth` needs `mesh_flag != 0` to trigger the locking depth zeroing.

Yeah, in practice, the only thing that `mesh_flag = 0` does is zero out the locking depth.  Renaming is a possibility but `zero_out_segment_locking_depth_because_this_segment_is_replaced_by_mesh` might be too long ;) 

> A property of the _mesh_ parameters (`use_mesh = True` or something) 

This is interesting.  My first question is how to deal with potential conflicts between the `*segment.csv` and `*mesh_parameters.json` files.  Consider this synthetic case:

-  `mesh_idx = 3` for a bunch of segments in a `*segment.csv` file.
-  `use_mesh = False` for the third in a `*mesh_parameters.json` file.

What would be a resolution in this case?  It seems to me that this issue would occur any time something like `use_mesh = False` were used in a `*mesh_parameters.json` file unless all relevant values of `mesh_idx` in a `*segment.csv` file were updated correctly.

jploveless:
I guess I was thinking that `use_mesh` would override the settings in the `*segment.csv`, essentially acting as a toggle between a mesh vs. rectangles. If `use_mesh = False`, then that mesh would be ignored later in the run, and the segments with `mesh_file_index = 3` therefore maintain their originally specified locking depths and are treated as rectangles. 

But I could see this getting nightmarish in terms of indexing, so if this ends up being something that becomes common (which I don't expect), it might be more straightforward to write a helper function that would pre-process a `*segment.csv` to adjust the `mesh_file_index` fields to reflect deletion of a mesh from a corresponding `*mesh.json`.

brendanjmeade:
> I guess I was thinking that `use_mesh` would override the settings in the `*segment.csv`, essentially acting as a toggle between a mesh vs. rectangles. If `use_mesh = False`, then that mesh would be ignored later in the run, and the segments with `mesh_file_index = 3` therefore maintain their originally specified locking depths and are treated as rectangles.

   I understand this now.


> But I could see this getting nightmarish in terms of indexing, so if this ends up being something that becomes common (which I don't expect), it might be more straightforward to write a helper function that would pre-process a `*segment.csv` to adjust the `mesh_file_index` fields to reflect deletion of a mesh from a corresponding `*mesh.json`.

   I think that's a very good solution.  In fact, I think a lot of real work is in this type of segment and mesh file pre-processing.

    Ready to merge this PR?

PR #227, opened 2025-08-05
feat: add missing accessors Estimation.tde_*_rate

### LINKED ISSUE #218: Standard matplotlib notebook
@aseyboldt Can you please take a look at: https://github.com/brendanjmeade/celeri/blob/main/notebooks/doc_matplotlib_vis.ipynb

I had this in a PR but somehow merged it already.   `¯\_(ツ)_/¯ `

aseyboldt:
Properties like `Estimation.tde_strike_slip_rates_kinematic` were only TODO stubs even before the refactor, and still only returned zeros.
I replaced them with actual implementations, and added separate ones for the smoothed kinematic rates. The properties also return dict with mesh indices as keys now, instead of the concatenated values, because I think we only ever want to use values that are specific to a particular mesh. That avoids having to first concatenate and then split them again (and possibly introducing errors along the way).

@brendanjmeade I think this unblocks and fixes #218, but let me know if there are further issues with it.

### DIFF
--- a/celeri/operators.py
+++ b/celeri/operators.py
@@ -2,7 +2,7 @@ import hashlib
 from dataclasses import dataclass, field
 from functools import cached_property
 from pathlib import Path
-from typing import Any
+from typing import Any, overload
 
 import addict
 import h5py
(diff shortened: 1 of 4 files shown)


brendanjmeade:
@aseyboldt Your edits added all of the fields that I was looking for to `estimation`.  Thanks!  I like the integer dictionary for accessing them.  It's much more natural.  I've updated the visualization notebook in PR https://github.com/brendanjmeade/celeri/pull/230.

There is one outstanding issue in the notebook and that's the plotting of the iterative convergence.  In the last cell of the notebook we have:

```
celeri.plot_iterative_convergence(estimation)
```

and that throws the error:

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[27], line 1
----> 1 celeri.plot_iterative_convergence(estimation)

File ~/Desktop/celeri-org/celeri/celeri/optimize_sqp.py:535, in plot_iterative_convergence(estimation, plot_in_bounds)
    533 meshes = estimation.model.meshes
    534 n_oob_vec = estimation.n_out_of_bounds_trace
--> 535 n_meshes, n_iter = n_oob_vec.shape
    537 assert len(meshes) >= n_meshes
    538 mesh_names = [mesh.name for mesh in meshes[:n_meshes]]

ValueError: not enough values to unpack (expected 2, got 1)
```
Can you please look into this and provide a fix?

aseyboldt:
Ah, is that an estimation from solve_sqp2? I think we never tested the plot_iterative_convergence function with those. I'll have a look.

brendanjmeade:
@aseyboldt What I'm looking for is a way to show the progress of the convergence of the iterative process both on individual meshes and overall.  Plotting something like the number of OOB elements at each iteration step.  I stored this in `n_oob_vec` in the original version, but don't know where to find it or similar in the modern.

brendanjmeade:
@aseyboldt I now see `estimation.n_out_of_bounds_trace` which gives the convergence with iteration number.  Nice.

The overall goal is to plot the % of coupling constraints satisfied (or not satisfied) on each mesh with each iteration step.  It's a little tricky with the logic, as some elements have coupling constraints and some have slip rate constraints.  We only want to see the convergence percentage for those mesh elements with coupling constraints.

I think you can help with that.

PR #233, opened 2025-08-08
new_coastlines

brendanjmeade:
Coastline and land data now accessed using cartopy.  WorldHiVectors.mat removes.  Faster as a side effect.


maresb:
I'd recommend something like the following (hopefully I didn't make a mistake):

```bash
git fetch origin
git checkout new_coastlines
git rebase main  # This will stop on a conflict with pixi.toml
git checkout --theirs pixi.lock  # Retain the original pixi.lock from `main`
git add pixi.lock  # Stage the original `pixi.lock` file
git rebase --continue
```

Then make a new commit at the end to update the lockfile

```bash
pixi install  # Install while making minimal changes to the environment, and update lockfile accordingly
git add pixi.lock
git commit -m "Relock dependencies"
git push --force  # Overwrite the remote branch
```

maresb:
@brendanjmeade, I mentioned that it's way easier to rebase where you just drop a commit.

In that case, my go-to workflow is `git rebase -i main` where `-i` is for "interactive".

This opens up an editor with one line per commit, in chronological order, first-top / last-bottom.

In the interactive rebase editor, all you need to do is delete the line you want to drop. (Making the line empty is also sufficient, since blank lines and comments are ignored. Alternatively, replace the leading `pick` action by `drop` or `d`.)

**In case anything goes awry**, delete all non-comment lines (or also comment lines) and save. An interactive rebase with zero actions gets interpreted as an abort. Even then if something goes wrong, you can run `git reset --hard ORIG_HEAD` immediately after a rebase to undo it. If you've been committing or rebasing after the rebase you want to undo, it's still possible but requires some advanced fishing around in Git's "reflog".

maresb:
@brendanjmeade, one last pixi trick:

In case you think your environment is corrupt, you can run:

```
pixi clean  # delete the current environments in the .pixi/ cache
pixi install  # reinstall, now from scratch, rather than only-if-necessary
```

PR #235, opened 2025-08-11
Remove okada_wrapper, take 2

maresb:
Rebased on `main`

TODO: why is cutde so much slower than okada_wrapper when I switch `get_okada_displacements` from `triangulation = "okada"= to `triangulation = "auto"`?


maresb:
As for validation, here's a comparison between this branch and some very old values from Western North America, all within the expected $10^{-12}$:
image
image

maresb:
I verified also for Japan. That was complicated due to the lightsabers, but it looks like my [auto-triangulation algorithm](https://github.com/maresb/celeri/blob/ff48510872eba10336a0522d7c1f598671d07016/celeri/okada/auto-triangulation.md) is working.

maresb:
Feel free to merge this if you feel satisfied. I'll be working on the dependency audit.

maresb:
Thanks so much @jploveless for the kind words! :blush: It would be nice to eventually port that out into cutde or a separate package. I'm also really glad that it fixes the "lightsaber" artifacts present in okada-wrapper.

brendanjmeade:
> @maresb, thank you for this implementation and for so clearly laying out the approach in `auto-triangulation.md`! I feel like that's such an important document for `celeri` but also for the much larger user community of dislocation elements in general.

Agree with @jploveless that `auto-triangulation.md` is elite.  Clear, informative about a complex topic.  Awesome!

maresb:
> @maresb I have a naive question: Where in this PR is the new cutde calcuation called in celeri?

@brendanjmeade, both `celeri.plot:plot_segment_displacements` and `celeri.spatial:get_segment_station_operator_okada` call `celeri.spatial:get_okada_displacements` which calls `celeri.okada.cutde_okada:dc3dwrapper_cutde_disp` which calls `cutde.halfspace:disp` imported via `from cutde.halfspace import disp as disp_cutde`. 

The key change in the codepath is in `celeri.spatial` to replace
 
```python
_, u, _ = okada_wrapper.dc3dwrapper(...)
```

 with

```python
u = dc3dwrapper_cutde_disp(...)
```

For reviewing this, you may find it easier to view the individual commits rather than all the changes at once. (I'll often mass-open one tab per commit and close tabs as I progress through a review.) I like to think that I'm usually fairly good about making a single logical change per commit, although I got a bit sloppy in this respect during my last few cleanup commits.

brendanjmeade:
>  TODO 2: it would be great to undo the coordinate transforms (I think this can be deferred) 

Interesting!  Which coordinate transforms are you thinking of here?  Local rotations to align with x-axis?  Map projections (pretty difficult to do without)?  Something else?

maresb:
Rebased on main with no other changes.

brendanjmeade:
> Primarily that. Like what we recently discussed last time. But for now I've satisfied the primary requirement of eliminating the okada_wrapper dependency. Once we get this and #237 merged, I can create the conda-forge package. So I'd prefer to unwind the x-axis alignment part in a follow-up.

Sounds good.


> Hmm, I haven't thought about this yet, but I don't see why we couldn't make versions of okada and cutde with the map projections built in. Unless that would offend your flat-earth sensibilities 😛. To be specific, I'm guessing that you use a fixed projection for the whole region, but alternatively you could do a separate projection around each station, putting each station at the center of its own E/N/up coordinate system. But I could see how this might be undesirably complicated since it changes methodology. (Or maybe this is how you already do it? I haven't checked.)

Let's _**not**_ pursue this.

PR #237, opened 2025-08-12
Audit pyproject.toml dependencies

maresb:
Based on #235, so please ignore all but these last few commits:
image

Once #235 is merged with okada-wrapper removed, I want to cut a new PyPI release to use as a basis for the conda-forge package. Accordingly I need to decide exactly what needs to be installed with this new version when someone calls `pip install celeri`.

My proposal is:

https://github.com/brendanjmeade/celeri/blob/05c0f5886007e5ec793ff99fa6a54cd5b428ba63/pyproject.toml#L14-L36

This:
* removes `ismember` as unused
* removes `pytest` since it's a development rather than a core dependency
* conditionally imports `ipython` in `celeri_solve.py` and removes it as a core dependency since it's a notebook dependency (I'm envisioning that some people might want to run celeri headless)
* adds the previously-missing dependencies `cartopy`, `colorcet`, `cvxpy`, `pydantic`, `shapely`, `xarray`, `zarr`
* does not include `polars`, `seaborn`, `matplotlib-inline`, `okada-wrapper`, even though they're referenced in notebooks
* does not include `papermill` even though it's used in `run_notebooks.py`
* assuming that they will be included shortly, I've for now left out `pymc` and `nutpie` for now, but we can include them in a follow-up release

These decisions are very subjective, so I'm happy to change things around if anyone has alternative preferences. I think of the `pixi.toml` as the batteries-included definition. I figure that once we get this published to conda-forge, people can define their own `pixi.toml` or `environment.yml` with their favorite Jupyter backend and plotting libraries.


brendanjmeade:
Clean up is good!


> * removes `ismember` as unused

Reasonable

> * removes `pytest` since it's a development rather than a core dependency

(Comment #1) I'm not sure about this about this because, as you mention, it's for development, and I run it locally with some frequency.  


> * conditionally imports `ipython` in `celeri_solve.py` and removes it as a core dependency since it's a notebook dependency (I'm envisioning that some people might want to run celeri headless)

Reasonable.

> * adds the previously-missing dependencies `cartopy`, `colorcet`, `cvxpy`, `pydantic`, `shapely`, `xarray`, `zarr`

Great!


> * does not include `polars`, `seaborn`, `matplotlib-inline`, `okada-wrapper`, even though they're referenced in notebooks

Reasonable.


> * does not include `papermill` even though it's used in `run_notebooks.py`

See (Comment #1) above.


> * assuming that they will be included shortly, I've for now left out `pymc` and `nutpie` for now, but we can include them in a follow-up release

Getting there!


> These decisions are very subjective, so I'm happy to change things around if anyone has alternative preferences. I think of the `pixi.toml` as the batteries-included definition. I figure that once we get this published to conda-forge, people can define their own `pixi.toml` or `environment.yml` with their favorite Jupyter backend and plotting libraries.

It's going to be great!

I do want to understand how we continue to develop with a "deployed" package.  Obviously, we're still adding things to celeri, but we'd like to run it as a deployed package as well, so we can get used to completely separating the code from data and have all the scripts callable system-wide.  Let's discuss this and get me educated!

maresb:
To elaborate my thinking, regarding comment 1 as it applies to `pytest` and `papermill`:

The primary audience of the PyPI package as defined by `pyproject.toml` is users who will run `pip install celeri` or `conda install celeri`. As developers, we definitely always want to work in an environment with `pytest` and `papermill` e.g. the one created by pixi. If we declare `pytest` in `pyproject.toml`'s `project.dependencies`, we're effectively saying that `pytest` is required for the core functionality of `celeri`, and that your environment is inconsistent if you don't have it.

To reorient, perhaps the main question here is if/how we declare our development dependencies. It's possible to declare a `dev` "extra" so that you could run `pip install celeri[dev]`. However, I think it's easier to instead do development with conda/pixi, and there's not really any good conda equivalent of `celeri[dev]`. Hence my preference to provide a batteries-included `pixi.toml` / `environment.yml` for both development and general usage that will always include all useful dependencies, and then we can keep the core package minimal.

brendanjmeade:
@maresb Thanks for clearly explaining your thoughts.  Let's discuss this today.

One guiding perspective:  Me and my collaborators will be the primary users and don't mind a few extra libraries if that makes our lives easier.  Practical convenience is more important than a lean pixi.toml/environment.yml

maresb:
Now:
* removes `ismember` as unused
* conditionally imports `ipython` in `celeri_solve.py` (I'm envisioning that some people might want to run celeri headless)
* adds the previously-missing dependencies `cartopy`, `colorcet`, `cvxpy`, `pydantic`, `shapely`, `xarray`, `zarr`
* does not include `polars`, `seaborn`, `matplotlib-inline`, `okada-wrapper`, even though they're referenced in notebooks
* keeps `pytest` and `ipython`, and adds `papermill` and `ipykernel` for convenience
* assuming that they will be included shortly, I've for now left out `pymc` and `nutpie` for now, but we can include them in a follow-up release

Issue #176, opened 2025-06-14
Making forward model predictions of velocities from prior models


brendanjmeade:
After a model is run, we may want to make predictions on other data given the state estimated from the prior run.  Here's how a new script named `celeri_predict.py` could do this.

Proposed command-line syntax:
```
celeri_predict new_config.json prior_output.pkl 
```

- `new_config.json`: Input files for new run.  Model geometry must be identical. `*station.csv` should have the locations where you'd like to make velocity predictions.
- `prior_output.pkl`: Output from prior run, including `Estimation` object, because we need the state estimate from a prior run.

**The general idea for how `celeri_predict` would work is very similar to `celeri_solve`**:
- Read `new_config.json file and process to create `Model` object.
- Instead of doing a solve, load the state estimate from the prior run (`estimation.state_vector`).  If the model geometry and constraints are unchanged from the prior run, then the `operator` matrix will have the same number of columns (but probably not rows) as the prior model.
- Make predictions with `operator` matrix from new data and `estimation.state_vector` from prior model.
- Write output as normal.

TODOs:
- [ ] Make sure the `Model` object is saved to `output.pkl` file.  Use cloudpickle to ensure that objects can be saved.
- [ ] Start building with a copy of `celeri_solve`.

maresb:
The title is missing a word

PR #238, opened 2025-08-13
Celeri forward calculations

brendanjmeade:
@jploveless.

This addresses issue: https://github.com/brendanjmeade/celeri/issues/176

Two new scripts:

1. `celeri_forward.py`.  Makes forward model predictions based on a prior `celeri_solve.py` solution.  Call as:
```bash
celeri_forward.py <path to output folder> <*_station.csv file for forward model precitions>
```
This is batched, too.  It runs 1000 stations at a time, so that it never creates large matrices.

2.  If you want to run  `celeri_forward.py`, you probably want some gridded locations for model evaluation.  That's what `create_grid_station.py` is for:  Call as:

```bash
create_grid_station.py <lon_min> <lat_min> <lon_max> <lat_max> --n_points=<number of grid points>
```

- `lon_min`: Minimum longitude
- `lat_min`:  Minimum latitude
- `lon_max`:  Maximum longitude
- `lat_max`:  Maximum latitude
- `--n_points=<number of grid points>`: Optional.  The default value is 100.

A little visualization because this is going to make awesome figures.
Screenshot 2025-08-12 at 10 42 47 PM

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -6026,8 +6026,8 @@ packages:
   timestamp: 1752907619396
 - pypi: .
   name: celeri
-  version: 0.0.2.post1.dev375+g01fd64516
-  sha256: b328526ac3338951e6855d3916b1eebef0090a7e3d824e79f8c34a630b38bd89
+  version: 0.0.2.post1.dev377+g7e4821d.d20250813
+  sha256: 5c3c8906907f066210fc7312c2ccddbb1d8a281806e69e7192f06013c83a97ff
   requires_dist:
   - addict
   - cutde


jploveless:
This is really nice! That contour figure is very motivating! 

Quick question: the output from both of these scripts is written to a filename based on a UUID. Can you clarify the reason for this, rather than something like `model_run_dir` + `station_file` + `"_forward.csv"` (where `station_file` might be `lon_min` + `lon_max` + `lat_min` + `lat_max` + `n_points` + `".csv"` or just `japan_100.csv`)?

Issue #142, opened 2025-04-11
Where to store the `runs/` directory when celeri is installed as a package?


maresb:
Currently the `runs` directory is located in `./../runs`. It's not particularly stable to make it relative to the current directory.

One idea would be to put it under:

```python
import platformdirs

runs_dir = platformdirs.user_cache_path("celeri") / "runs"
print(f"{runs_dir}")  # ~/.cache/celeri/runs
```

And then we would emit a log message containing this location so that the user knows where to find it.

Any thoughts or counterproposals?

maresb:
@brendanjmeade, @aseyboldt, in order to finish out #139 I could use some feedback here.

If we are running in "celeri development mode" then it makes sense to put the `runs/` directory relative to the current directory. But now that we're aiming to offer `celeri` as an installable library this no longer makes sense.

Is the platform-dependent `celeri` cache directory a good choice?

aseyboldt:
Could we just make the path a config option in the commands.json?

brendanjmeade:
Currently, the path for the output directory is specified in `*_command.json` files as:

```
"base_runs_folder": "../runs/"
```

See: https://github.com/brendanjmeade/celeri/blob/969d339422c539a7ed704f5542c33ac08048ab44/data/command/japan_command_cmi.json#L4 for an example

I agree with @maresb that not doing this as a relative path is the right thing to do.

I think it reasonable and recommend that we follow @aseyboldt's suggestion that we should ask users to specify a path in the command file using the same field that they do now, just not with a relative path.

This is another nice step to separate code from output!

maresb:
Thanks a lot for the discussion!

Requiring an absolute path in the command config is a great way to ensure that a config is usable only by a single user, i.e. `"base_runs_folder": "/home/my-name-not-yours/celeri/runs"`.

So now what? To me this indicates a potential need for a separation between user-level config and project-level config. Unfortunately having multiple configurations risks introducing a lot of complexity in the sense of hidden state. Generally speaking, it's really easy to set something but then forget it some weeks or months later, and then not understand why things are behaving in a non-standard way. And this isn't really only about `base_runs_folder` but about more general strategy.

A few potential approaches:
1. Configuration is difficult, so try and avoid it as much as possible, and make opinionated decisions, like maybe put all runs in a celeri cache directory.
2. Have opinionated defaults, but allow the user to create config files to override those defaults. Remind the user verbosely of the config file locations whenever one is used.
[...]

PR #253, opened 2025-08-21
Clarify config file name

maresb:
I was a bit confused at first by the original description in https://github.com/brendanjmeade/celeri/issues/142#issuecomment-3207531306, so I reworded it to better explain the usage.

I also made a few style changes like switching attribute descriptions to from comments to docstring format because this enables hover hints in VS Code.


brendanjmeade:
This is very helpful.  Thanks!

PR #254, opened 2025-08-21
Add mcmc solver for uncertainty quantification

aseyboldt:


brendanjmeade:
@aseyboldt @maresb Is the blocker here just pixi.lock?

brendanjmeade:
@aseyboldt Great to see the elastic velocity constraints land.  QQ: Does the MCMC solver also support fault slip rate constraints in the same way that the SQP solver does?  I'm checking because we're looking for 1:1 feature parity here.

aseyboldt:
@brendanjmeade I think this is ready for review.

It now includes support for the velocity constraints, a small example notebook, and also fixes the `out_of_bounds` trace in the solve_sqp2 (would have been cleaner as a separate PR...).

It does *not* support fault slip rate constraints as the SQP/SQP2 solver does however. I think the best way to add that would be as a soft constraint, instead of a strict one. But I think that requires quite a bit more experimentation and changes the the configuration, and I'd rather do that as a potential follow-up project item.

I'm working on the little write-up we mentioned during the last call, that's still coming up.

brendanjmeade:
@aseyboldt I've merged this and will continue to test it.

My first question is, how do you plot the SQP convergence history? `celeri.plot_iterative_convergence(estimation)` still fails as the last cell in `doc_matplotlib_vis.ipynb`

aseyboldt:
Did you rerun the estimation? An old estimation doesn't work, because the `n_out_of_bound_trace` changed.
I get this plot:
image

brendanjmeade:
That looks great!

I just tried running a new celer_solve and hit: https://github.com/brendanjmeade/celeri/issues/259

PR #269, opened 2025-09-21
Use the eigenvalues to impose the standard Matérn prior for the GP

maresb:
I'm not sure if it makes a difference, but we should be attenuating the coefficients in our GP according to the standard decay for the Matérn prior. @aseyboldt, for some reason this is running really slowly for me (also with `main`). Could you check if this works for you?

### DIFF
--- a/celeri/operators.py
+++ b/celeri/operators.py
@@ -287,6 +287,7 @@ class TdeOperators:
 @dataclass
 class EigenOperators:
     eigenvectors_to_tde_slip: ByMesh[np.ndarray]
+    eigenvalues: ByMesh[np.ndarray]
     eigen_to_velocities: ByMesh[np.ndarray]
     eigen_to_tde_bcs: ByMesh[np.ndarray]
     linear_gaussian_smoothing: ByMesh[np.ndarray]
@@ -480,6 +481,7 @@ class _OperatorBuilder:
     mogi_to_velocities: np.ndarray | None = None
     slip_rate_to_okada_to_velocities: np.ndarray | None = None
     eigenvectors_to_tde_slip: dict[int, np.ndarray] = field(default_factory=dict)
+    eigenvalues: dict[int, np.ndarray] = field(default_factory=dict)
     rotation_to_tri_slip_rate: dict[int, np.ndarray] = field(default_factory=dict)
     linear_gaussian_smoothing: dict[int, np.ndarray] = field(default_factory=dict)
     tde_to_velocities: dict[int, np.ndarray] = field(default_factory=dict)
@@ -546,9 +548,11 @@ class _OperatorBuilder:
         assert self.eigen_to_velocities is not None
         assert self.eigen_to_tde_bcs is not None
         assert self.eigenvectors_to_tde_slip is not None
+        assert self.eigenvalues is not None
 
         eigen = EigenOperators(
             eigenvectors_to_tde_slip=self.eigenvectors_to_tde_slip,
+            eigenvalues=self.eigenvalues,
             linear_gaussian_smoothing=self.linear_gaussian_smoothing,
             eigen_to_velocities=self.eigen_to_velocities,
             eigen_to_tde_bcs=self.eigen_to_tde_bcs,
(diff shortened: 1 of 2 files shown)


aseyboldt:
I'm not sure yet, but to me that looks more like a bug than an actual
difference due to the prior. How could the model possibly know that there
are definitely no small scale variations?
I haven't spotted a concrete problem in the code, but could it be related
to how the eigenvectors are stored in celeri? Maybe we somehow ended up
treating the eigenvectors of the strike slip component as higher frequency
components and scale them down more?

brendanjmeade:
@maresb and @aseyboldt It took me a while to read through this.  Investigations here are smart so that we continue to bake in nice properties to the solution to improve the cavelier welded-together approach that I brought.

> * The covariance kernel, e.g. "Matérn" or "squared exponential" (the smoothness $\nu\to\infty$ limit of Matérn). Celeri currently uses Matérn with $\nu=\tfrac{1}{2}$ which is probably not as smooth as desired, and we made the tentative decision to switch to "squared exponential". The kernel has two more parameters: the amplitude scale $\sigma$ and the correlation length scale $\ell$ which is a distance between two mesh elements.

I'd like to understand the relative merits of these. Is there a theory that suggests the appropriateness of one form or another?

> * The triangulation. There are two main issues here:
>    1. There's yet another third distance scale for the resolution of the mesh: the distribution of distances between adjacent centroids
>    2. A weight must be chosen for each triangle. Currently celeri uses a constant weight, and this has the advantage of providing more detail along refined regions. The more geometrically principled approach is to weight by area, which plays nicely with continuum limits. To get the best of both worlds, we could default to weighting by area, and then when desired, have an optional "measure" factor on top of the area for detail control.

Very thoughtful and reasonable.

> In addition to the above-mentioned distance-based kernels, there are also neighbor-based Laplacian kernels. One of the wishes was to check "distance-along-the-mesh" kernels, and the Laplacians are effectively both the best and easiest way to achieve this. (The naive distance-along-the-mesh kernels are not even positive-definite!!!) It's pretty straightforward to trade out the KL modes for the Laplacian modes, so this provides another "model selection" check.

I'm certainly open-minded about this.  I found distance weighted eigenmodes, and they just seemed to work, so I went with it.  Wise people have warned me off the geodesic distance!

> When it comes to modeling the coupling constant, we enforce the constraint with a logit transform, and thus in logit space we need to set the amplitude scale, and also the mean! (Usually when discussing GPs we set the mean to zero.) The amplitude scale in logit space determines how sharply we transition between the 0 and 1 extremes. The mean determines our default guess for the coupling. Based on previous discussion, rather than using mean 0 in logit space, corresponding to default coupling of 0.5, we want the default coupling to be closer to 0.8. Also, we want to choose the amplitude so that coupling transitions gradually rather than sharply.

I agree with you on 0.8 as starting point.  I'm less sure about how to tune the transitions.

brendanjmeade (commented):
Matern is happening!

elc45:
@brendanjmeade  The modes don't depend on the prior; we're just changing the prior on the coefficients $\alpha_i$ by which we multiply each mode in the sum $\sum_i{\alpha_i u_i}$ being plotted above. Each subplot in the first plot is generated by drawing $\alpha_i$ from $N(0, 10)$ (then multiplying by $u_i$) for $i$ in 1 to $n$. The subplots in the second plot draw $\alpha_i$ from $N(0,\lambda_i)$ for each $i$ in $n$. In both, the set $u_i$ are the same top $n$ modes of the covariance matrix. The crucial difference is in the behavior of the sum as $n$ grows.

The actual parameters in the model are these $\alpha_i$ values, and hence what are given priors and what inference is done over. In the scope of this PR, the eigenmodes can be considered attributes inherent to the mesh. If we shift to Matern 5/2, changing the covariance kernel, these eigenmodes _will_ change, but just once.

elc45:
@brendanjmeade the eigenmodes are just the eigenvectors of the covariance matrix informed by the pairwise-distances between tde centroids, so $\lambda_i$ are the eigenvalues of each of those eigenvectors.

elc45:
@brendanjmeade each base eigenmode $u_i$ is indeed only multiplied by a scalar $\alpha_i$. Each of those subplots is a different weighted sum of _all of the n eigenmodes together_, with the weights being random draws of $\alpha$. So those plots above aren't showing different eigenmodes, they're showing different weighted sums of all of them. Does that make sense?

Issue #259, opened 2025-08-26
divide by zero encountered in matmul


brendanjmeade:
@aseyboldt @maresb 

The last merged PR (https://github.com/brendanjmeade/celeri/pull/254) has introduced an array of errors that are clearly exhibited in a celeri_forward.py run.

Fixing this is an immediate priority.  It's also important to understand why testing didn't catch this.

```python
celeri on  main [$?] via 🐍 v3.13.5 via 🧚 v0.46.0 (default)
❯ python celeri/scripts/celeri_solve.py data/config/japan_config.json
INFO       RUN_NAME: 0000000060
           - celeri.celeri_util:get_logger:154 - 2025-08-26 09:39:55
INFO       Write log file: /Users/meade/Desktop/celeri-org/celeri/runs/0000000060/0000000060.log
           - celeri.celeri_util:get_logger:155 - 2025-08-26 09:39:55
INFO       Reading data files
           - celeri.model:read_data:101 - 2025-08-26 09:39:55
SUCCESS    Read: /Users/meade/Desktop/celeri-org/celeri/data/segment/japan_segment.csv
           - celeri.model:read_data:114 - 2025-08-26 09:39:55
SUCCESS    Read: /Users/meade/Desktop/celeri-org/celeri/data/block/japan_block.csv
           - celeri.model:read_data:119 - 2025-08-26 09:39:55
SUCCESS    Read: /Users/meade/Desktop/celeri-org/celeri/data/mesh/nankai.msh
           - celeri.mesh:from_params:497 - 2025-08-26 09:39:55
SUCCESS    Read: /Users/meade/Desktop/celeri-org/celeri/data/mesh/japan.msh
           - celeri.mesh:from_params:497 - 2025-08-26 09:39:55
SUCCESS    Read: /Users/meade/Desktop/celeri-org/celeri/data/mesh/sagami.msh
           - celeri.mesh:from_params:497 - 2025-08-26 09:39:55
SUCCESS    Read: /Users/meade/Desktop/celeri-org/celeri/data/mesh/japan_mock_cmi.msh
           - celeri.mesh:from_params:497 - 2025-08-26 09:39:55
SUCCESS    Read: /Users/meade/Desktop/celeri-org/celeri/data/station/japan_station.csv
           - celeri.model:read_data:156 - 2025-08-26 09:39:55
SUCCESS    Read: /Users/meade/Desktop/celeri-org/celeri/data/mogi/japan_mogi.csv
[...]

PR #273, opened 2025-09-26
Update numba 0.61.2 → 0.62.0 and numpy 2.2.6 → 2.3.3

maresb:
Also llvmlite 0.44.0 → 0.45.0.

This is meant to resolve https://github.com/brendanjmeade/celeri/issues/259


maresb:
@brendanjmeade, when you get the chance, could you please check to see if this gets rid of those warnings on your mac?

maresb:
The `pixi update` was unnecessary. This PR was a targeted update of only `numpy` and `numba` via running `pixi update numpy numba`. On the other hand, `pixi update` updates all dependencies, leading to more moving parts. But I just verified, and `pixi update` doesn't downgrade `numpy` or `numba`, so I'm pretty baffled why this didn't fix the problem.

Note that the error is pretty nonsensical: `divide by zero encountered in matmul`. I'm no expert on obscure matmul implementations, but AFAIK, division is not typical.

For details on the root cause, see https://github.com/numpy/numpy/issues/28687. There it's indicated that this is a Mac problem.

Since this PR is merged, let's continue discussion at #259.

PR #284, opened 2025-10-29
Add segment slip rate regularization

aseyboldt:
This adds the slip rate regularization term we discussed on monday.

This pushes all slip rates at the segments a little to zero, how much is controlled by the new `segment_slip_rate_regularization` parameter.

I've been experimenting with different values on the new wna example *without any hard slip rate constraints*.

0.01 seems clearly too small, it only regularizes the values slightly. 10 is clearly too big, it makes the overall fit much worse. I think at least for this dataset, 0.1 or 1 are decent choices.

Here are the resulting residual mean squared errors with different values for the regularization:

10: 20 (mm/yr)^2
1: 13 (mm/yr)^2
0.1: 11.8 (mm/yr)^2
0.01: 11.4 (mm/yr)^2


aseyboldt:
I'll put the other solver flags in a separate PR.

This should already work with constraints as well, I just did the testing to find decent values for the regularization parameter without those.

aseyboldt:
I added the commits for the soft constraints both in sqp2 and mcmc.
I also added config items for all of them, and included some documentation about those values. If the docs don't clear up things, please let me know and I'll explain it a bit more.

I also included a new config item for the sqp2 objective. The more stable one is `norm2`. But I don't think this really saves using hard constraints everywhere. While it does find a valid point, the resulting fit is horrible.

celeri/config.py
```diff
@@ -2,12 +2,21 @@
...
+    "expanded_norm2",
+    "sum_of_squares",
+    "qr_sum_of_squares",
+    "svd_sum_of_squares",
+    "norm2",
+    "norm1",
+]
+
 
 class Config(BaseModel):
```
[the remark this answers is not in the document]
brendanjmeade: @aseyboldt Understood.  Once this PR is ready, I'll add these to the config files so that they are explicitly exposed.  Important question:  Which specific `Sqp2Objective` setting led to the solver completing with the additional constraint?  It's important to document this setting.  You mentioned this during our chat two weeks ago.

celeri/config.py
```diff
@@ -80,6 +80,41 @@ class Config(BaseModel):
...
+    """Weight for enforcing slip rate bounds at segments.
+
+    This is used in `solve_sqp2` to enforce slip rate bounds softly.
+    """
+
+    segment_slip_rate_hard_bounds: bool = True
+    """Enforce hard slip rate bounds at segments.
+
+    This should be disabled when using soft slip rate bounds
+    via `segment_slip_rate_bound_sigma`.
```
maresb: But this is `True` and `segment_slip_rate_bound_sigma` is `3.0`. :thinking:

Why is this the case, and should it be enforced in the code?

celeri/config.py
```diff
@@ -80,6 +80,41 @@ class Config(BaseModel):
...
+
+    The mcmc solver does not support hard bounds.
```
maresb: ...and will raise an error if this is true.

TODO: test

celeri/config.py
```diff
@@ -80,6 +80,41 @@ class Config(BaseModel):
     Reasonable values might be in the range of 0.01 to 10.
     """
 
+    segment_slip_rate_bound_weight: float = 100.0
+    """Weight for enforcing slip rate bounds at segments.
+
+    This is used in `solve_sqp2` to enforce slip rate bounds softly.
+    """
+
+    segment_slip_rate_hard_bounds: bool = True
```
maresb: How about `segment_slip_rate_bounds_type: Literal["hard", "soft"] = "hard"`?

I find the bool confusing.

celeri/solve_mcmc.py
```diff
@@ -315,6 +323,109 @@ def _build_pymc_model(model: Model, operators: Operators) -> PymcModel:
...
+                    sigma=observed_sigs,
+                    observed=observed_rates,
+                )
+
+        # model.segment.ss_rate_bound_flag (and ds_rate_bound_flag, ts_rate_bound_flag)
+        # indicates if we have an interval bound for that component. The lower and upper
+        # bounds are stored in model.segment.ss_rate_lower_min and
+        # model.segment.ss_rate_upper_max. We use two pm.Bound(pm.Normal) likelihoods
+        # to implement the interval constraint. The standard deviation of the bound
+        # is set to model.config.segment_slip_rate_bound_sig.
```
maresb: sig -> sigma

celeri/solve_mcmc.py
```diff
@@ -337,6 +448,12 @@ def solve_mcmc(
 
     import nutpie
 
+    if model.config.segment_slip_rate_hard_bounds:
+        raise ValueError(
+            "Hard bounds on segment slip rates are not supported in MCMC solve. "
+            "Please use soft bounds with `segment_slip_rate_bound_sigma` instead."
```
maresb: Please change `segment_slip_rate_bounds_type` to `soft` and set a value for `segment_slip_rate_bound_sigma`

PR #288, opened 2025-11-11
Switch pixi.toml from [project] to [workspace]

maresb:
This is a super-minor maintenance update.

If you encounter an error after this is merged, then you need to update pixi:

```bash
pixi self-update
```

Pixi has treated `pixi.toml` as a workspace manifest since 0.43.0, and as of 0.57.0 it explicitly deprecates the `[project]` table in favor of `[workspace]`, emitting warnings when the old name is used. Updating our manifest removes those warnings.

<https://pixi.sh/dev/CHANGELOG/#0430-2025-03-20>

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -2,7 +2,7 @@
 # It is equivalent to the `environment.yml`. The environment can be activated
 # with the command `pixi shell`.
 
-[project]
+[workspace]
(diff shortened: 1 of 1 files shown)


brendanjmeade:
Merged and `self-update` complete

PR #304, opened 2025-11-16
Plotting, hashing, and docs

elc45:
Changes include reorganizing plotting function imports, continued documentation additions, and modifying the logic for operator caching such that constrain changes which do not affect the geometry of the problem do not yield different hashes.


brendanjmeade (commented):
I'd like to understand https://github.com/brendanjmeade/celeri/pull/304/commits/b5ad51480407a8e7b3cf7b9fa6be5604c0179a85 better as a part of the general caching strategy when we have a chance to chat!

brendanjmeade (approved):
Thanks for both the comments and bringing back the hdf5 writing.  Please merge and I'll run it with the latest model.

elc45:
Will do, just need to fix up a few things with the hdf5 file saving (a few Types have changed here and there since the function was removed), and then compare it with the file you shared in the issue to make sure they're the exact same format.

.gitignore
```diff
@@ -3,6 +3,7 @@ celeri.egg-info/
 UNKNOWN.egg-info/
 output/
 runs/
+data/operators/*_segments.json
```
aseyboldt: I think this is not needed anymore?

PR #306, opened 2025-11-20
Require Python 3.13

maresb:
Since we have a limited userbase, I don't think we need to offer support for previous versions of Python. Let's just require the latest Python 3.13, and continue doing so on a rolling basis, in order to lighten our testing burden.

This should be helpful for #284


maresb:
Excellent @brendanjmeade! I'm about to run this by @aseyboldt in a few minutes to verify that we're aligned...

PR #310, opened 2025-11-24
EXPERIMENT: time series notebook

aseyboldt:
This isn't ready for review just yet, only for future reference!

The notebook contains a sketch about how we could add the time dependency into the mcmc solver using GP basis functions on (for instance) elastic tde slip rate eigen function parameters.


brendanjmeade:
@aseyboldt I propose we simply merge this for the sake of future experiments!

PR #311, opened 2025-11-24
Tests

elc45:
Draft for now. I need to figure out how to do the smart recompute test without needing to have all those csv files in the repo...

### DIFF
--- a/tests/test_solve_dense.py
+++ b/tests/test_solve_dense.py
@@ -2,7 +2,7 @@ import pytest
 import celeri
 import numpy as np
 
-@pytest.mark.array_compare(rtol=1e-7, atol=1e-9)
+@pytest.mark.array_compare(rtol=1e-4, atol=1e-9)
 @pytest.mark.parametrize(
     "config_name",
     ["test_japan_config", "test_wna_config"],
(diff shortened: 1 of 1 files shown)


brendanjmeade:
@elc45 I've skimmed through the assembly removal part.  I effectively thought about it backwards: What if we had assembly today?  Removing it is cleaner.  The big question: Is it giving the same answer on a WNA model?

elc45:
Removing `Assembly` (#316) is purely an implementation change / refactor, thus doesn't change any solutions.

brendanjmeade:
@elc45 @aseyboldt @maresb I definitely remember those tests being created. Are they still there?  Does this mean that we're running without tests against previous model results?  

 > Removing Assembly (https://github.com/brendanjmeade/celeri/pull/316) is purely an implementation change / refactor, thus doesn't change any solutions.
 
 That's the dream!  I just writing to check that this is the case?  Perhaps on Monday we can review the test suite?

aseyboldt:
I'm sure we discussed this before, but it looks like we never actually implemented it.
I'm not sure how that slipped through the cracks, maybe because I implemented similar tests in nutpie at the same time, I somehow got confused and thought I did it here as well?
I think we should consider this a left-over from the previous project, and just do it now.

brendanjmeade:
This PR is a lot, and I LGTM it in the spirit of:

151sj28e8yab1

elc45:
Haha, I was just telling Bernard this morning that it looks like a lot more code than it really is... He's going to take a look and review it, then will merge shortly!

maresb:
Superseded by #334

PR #323, opened 2025-12-02
Fix result_manager bug with MCMC solution

### LINKED ISSUE #317: MCMC solve output file compatibility with result manager
@aseyboldt @elc45 

When trying to run `result_manager` (https://github.com/brendanjmeade/result_manager) on the output from an MCMC solve, I get the error below. This is because meshes.csv is not being produced for the MCMC case.  I think this file should be produced with the mean elastic, coupling, and kinematic values populating the file.  Yes, this is admittedly less than what the full MCMC solution gives us, but it's a great way to quickly visualize the results interactively!

```
❯ result_manager
...
FileNotFoundError: [Errno 2] No such file or directory: '/Users/meade/Desktop/wna/runs/0000000167/model_meshes.csv'
...
```

elc45:
@brendanjmeade this should fix #317 ; I checked to make sure that the resulting run directory can now be opened with `results_manager`, but I couldn't do much spot checking since the GUI fries my little travel laptop... let me know if you see anything that doesn't look right- I'll be back home at my desktop tomorrow and happy to discuss further :)

### DIFF
--- a/celeri/solve.py
+++ b/celeri/solve.py
@@ -137,30 +137,49 @@ class Estimation:
     @cached_property
     def mesh_estimate(self) -> pd.DataFrame | None:
         """A dataframe containing the estimated slip rates and couplings for each mesh."""
-        if self.operators.tde is None or self.operators.eigen is None:
+        if self.operators.tde is None:
             return None
         if self.tde_strike_slip_rates is None or self.tde_dip_slip_rates is None:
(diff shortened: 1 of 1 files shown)


brendanjmeade:
Working now.  Much appreciated!  Screenshot 2025-12-01 at 10 15 28 PM

Issue #325, opened 2025-12-02
Caching issues?


brendanjmeade:
@elc45 I'm writing about two issues that may be related to the caching:

1. At the end of a celeri-solve run, I see the warning message below.  I _think(?)_ this is new.  Can you look into why this is happening?
```
/Users/meade/Desktop/celeri-org/celeri/.pixi/envs/default/lib/python3.13/site-packages/zarr/api/asynchronous.py:247: ZarrUserWarning: Consolidated metadata is currently not part in the Zarr format 3 specification. It may not be supported by other zarr implementations and may change in the future.
  warnings.warn(
```

2. This one is more subtle, and I'm pretty sure it's related to the caching of previous results.  Today I ran an experiment where I took an existing model and removed one tectonic block.  This involved deleting the relevant segments from a segment file and the internal block point from the block file.  The modified files can be found in a [WNA branch](https://github.com/brendanjmeade/wna/tree/remove_modify_small).  When running with the modified files, I get the error below related to matrix sizes.  However, when deleting all of the preexisting files in the operators folder (forcing a recompute of all matrices), the updated model runs just fine.  This makes me think that the caching is not working well in this case, where the updated model geometry removes segments.
   - This should be fixed either by updating the current implementation or reverting to the previous (and more coarse) implementation.  It's reasonable to consider both options here.
   - Related to this, it's important that there's a setting in the config file that allows us to.  This certainly used to exist, and its return would be appropriate as it would aid in situations where I want to force a recompute without deleting all of the cached files.

```INFO       | 2025-12-02 08:14:24 | celeri.celeri_util:get_logger:154 | RUN_NAME: 0000000174INFO       | 2025-12-02 08:14:24 | celeri.celeri_util:get_logger:155 | Write log file: /Users/meade/Desktop/wna/runs/0000000174/0000000174.log
INFO       | 2025-12-02 08:14:24 | celeri.model:read_data:113 | Reading data files
SUCCESS    | 2025-12-02 08:14:24 | celeri.model:read_data:126 | Read: /Users/meade/Desktop/wna/data/segment/graham_nshm23_merge_connected_fixed_segments_3_constraints_1_borderlands_constraint_fix_locking_depths_damped_segmesh.csv
SUCCESS    | 2025-12-02 08:14:24 | celeri.model:read_data:131 | Read: /Users/meade/Desktop/wna/data/block/wna_block.csv
[...]

PR #327, opened 2025-12-02
Caching fixes

elc45:
Fixes zarr warning in #325 and implements a `force_recompute` field in the config which ignores cached operators.


brendanjmeade:
@elc45 Looking good to me!

PR #339, opened 2025-12-09
Refinements and fixes of SQP2 documentation

maresb:
While looking at the code I realized that I was confused about some of the details about how the approximate coupling bounds are computed. The root of my confusion was the term "kinematic bounds" which is a minor misnomer. In fact the kinematic velocity alone is never constrained.

To fix this, I reworded things so that we refer more precisely to "kinematic anchor points" instead of "kinematic bounds". Correspondingly, I replaced my mistaken description of the trapezoidal region with an accurate description of the triangular region.

I also added some pretty intense ASCII art illustrations to the docstrings. (I know you're wondering, and I can attest that LLMs are NOT helpful when it comes to ASCII art. :joy:)

### DIFF
--- a/celeri/optimize.py
+++ b/celeri/optimize.py
@@ -184,7 +184,7 @@ class SlipRateLimitItem:
     kinematic_lower, kinematic_upper : np.ndarray | None
         Geometric anchor values along the kinematic axis used to construct
         the convex envelope of the bowtie constraint. These are NOT hard
-        bounds on the kinematic velocity; see `_tighten_coupling_bounds`
+        bounds on the kinematic velocity; see `_tighten_kinematic_anchors`
         for details.
     """
 


maresb:
> I recall that Adrian's version updated both the kinematic and elastic mesh velocities. Am I wrong in remembering this?

Adrian's version updates an evolving pair of linear inequalities in kinematic and elastic velocities.

> I haven't quite understood what the "kinematic anchor points" are now. Are these the evolving kinematic bounds?

They are geometric reference points for constructing the inequalities. If by "kinematic bounds" you mean something like `kinematic <= C` then *no*, and this is where I was confused.

Taking a step back, as we know, the goal is to start with two or more linear constraints that, in the beginning, allow all physically plausible values, and in the end allow all physically plausible values from one of the two wedges. For all practical purposes, what happens in between isn't particularly relevant as long as it works.

What I'm trying to do here is to gain a precise understanding of what happens in between. Since we went through many iterations, my mental model of "what happens in between" was outdated, and correspondingly what I previously wrote in the docstrings was incorrect.

In the end we managed to use only two constraints, so there are no trapezoids! Having just two linear inequalities in the plane means that we have evolving triangular wedges. The wedges are determined by the lines through geometrically-constructed anchor points.

"Kinematic bounds" makes me think of trapezoids. "Anchor points" makes me think of geometric constructions. This wording helps me to converge to the right geometric model.

It's tricky with words (and even with ASCII art) to coherently describe the geometric construction, so I'd be happy to take ~5 min on Monday to sketch it out with my tablet.

> That's interesting. I might have hope for that one!

Next big thing!!! Where's the VC and Euro innovation grants??? :money_mouth_face:

Issue #318, opened 2025-11-26
snap_segments tests


elc45:
Do we currently have any tests for `snap_segments.py`? I've changed some code in that file, but realized I don't have a way of testing it. If someone has some example usage, it'd be nice to add some tests for it to make sure I'm not breaking it!

brendanjmeade:
@jploveless is the creator of the very clever `snap_segments.py` so perhaps he has ideas?

PR #340, opened 2025-12-10
Improvements to tests

elc45:
- Add a new test file for snap_segments #318 
- Break up the caching tests into their own file, since it's a distinct thing.
- Make pytest fixtures for `assemble_and_solve_dense` in the operators tests to cut runtime in half
- Test that MCMC runs produce the files needed for `result_manager`
- Regroup/rename test actions

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,9 +1,0 @@
-import pytest
-
-
-def pytest_generate_tests(metafunc):
-    if "config_name" in metafunc.fixturenames:
-        metafunc.parametrize(
-            "config_name",
-            ["test_japan_config", "test_wna_config"],
-        )
--- a/tests/test_solve_dense.py
+++ b/tests/test_solve_dense.py
@@ -4,6 +4,7 @@ import numpy as np
 
 
 @pytest.mark.array_compare(rtol=1e-4, atol=1e-9)
+@pytest.mark.parametrize("config_name", ["test_japan_config", "test_wna_config"])
 def test_operator_tde_to_velocities(config_name):
     config_file = f"./tests/configs/{config_name}.json"
     config = celeri.get_config(config_file)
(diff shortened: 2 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - test-cleanup
```
aseyboldt: I guess this should be deleted now? Or what is this about?

tests/conftest.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+def pytest_generate_tests(metafunc):
+    if "config_name" in metafunc.fixturenames:
```
aseyboldt: Do we really need this level of magic? I'm not even sure what it does precisely...

tests/test_cache.py
```diff
@@ -0,0 +1,93 @@
...
+    4. Compares the selective recompute results with full recompute to ensure they match
+    
+    Args:
+        config_file: Path to the test configuration file
+    """
+    logger.disable("celeri")
+    cache_dir = None
+    try:
+        test_logger.info("Computing operator with original segment file")
+        model = celeri.build_model(config_file)
```
aseyboldt: I think it would be better to create a config object first, change the cache dir to a tempdir from `tempfile`, and create the model with that modified config. That way we control the directory directly, and can simplify this code a bit?

tests/test_output_files.py
```diff
@@ -52,81 +49,29 @@ def test_celeri_solve_creates_output_files(config_file):
...
+    """Test that celeri_solve_mcmc.py creates the HDF5 file and CSV files required by result_manager."""
+    config = celeri.get_config(config_file)
+    config.solve_type = "mcmc"
+    model = celeri.build_model(config)
+    for mesh in model.meshes:
+        if mesh.config.elastic_constraints_ss is not None:
+            mesh.config.elastic_constraints_ss = ScalarBound(lower=None, upper=None)
+        if mesh.config.elastic_constraints_ds is not None:
+            mesh.config.elastic_constraints_ds = ScalarBound(lower=None, upper=None)
+    estimation = celeri.solve_mcmc(model, sample_kwargs={"tune": 2, "draws": 2})
```
aseyboldt: Looks like running pre-commit or `ruff format` might be a good idea ;-)

PR #360, opened 2026-01-07
Vertical deformation option

elc45:


elc45:
I guess we could still sample the posterior predictive for the up velocities once we have the trace.

brendanjmeade:
@elc45 @aseyboldt There's a ton here.  I think the most practical thing to do is merge it and I'll start trying it out!

PR #367, opened 2026-01-13
Replace power-exponential kernel with Matérn

### LINKED ISSUE #362: Recommendation: remove support for power-exponential kernel
In my comment https://github.com/brendanjmeade/celeri/pull/356#discussion_r2665562548 I suggested dropping support of power-exponential in favor of Matérn. I wanted to elaborate my rationale here.

I finally managed to dig into analyzing what's the precise difference between the two, and the results indeed point strongly in favor of keeping only Matérn unless there is some historically-important reason for retaining power-exponential.

## The math

The power-exponential kernel is

$$k(d)=\exp(-(d/\ell)^p).$$

This is only positive-definite if $0<p\leq2$. (Not being positive-definite is very bad: it produces invalid covariance matrices, leading to situations where log-p *grows* quadratically in certain directions.)

<details>
<summary>proof outline</summary>
If $p>2$ then you can demonstrate negativity by example: choose 3 points in 1-D with domain $x\in\\\{-\epsilon,0,\epsilon\\\}$, and consider the function given by the values $y=(1,-2,1)$ over these points. The quadratic form evaluated on $y$ is $y^T K y$ and has leading order $(8-2^{p+1})\epsilon^p$, so for $p>2$ it represents a function of negative squared-length.

For $0<p\leq 2$, positive-definiteness follows from Schoenberg-type results.
∎
</details>

The $p=2$ case is special since it's the squared-exponential (Gaussian) kernel, which is identical to a Matérn kernel with $\nu=\infty$. For practical modeling, this choice tends to work poorly because it imposes unnaturally-high smoothness on functions.

In the remaining positive-definite range $0<p<2$, the power-exponential kernel behaves like a Matérn with $\nu=\tfrac{p}2$ that only reaches $\nu<1$. But at $p=2$ it jumps to being infinitely smooth (squared-exponential). 

Even though the kernels are not exactly the same, the power-exponential kernel is in precisely the same smoothness class as a Matérn kernel with smoothness parameter $\nu=\tfrac{p}{2}$. Where they overlap, they are practically interchangeable because the main qualitatively important properties of a kernel are length scale and smoothness; any remaining properties tend to be minor details.

## Conclusion

The power-exponential is basically equivalent to a subset of Matérn parameters, but this subset doesn't include the very common once- or twice-differentiable regime without jumping all the way to infinitely smooth, which is usually too smooth for robust modeling. In particular, my recommendation to use $\nu=\tfrac52$ would correspond to $p=5$, which is not possible with the power-exponential kernel (because it fails to be positive-definite).

## Recommendation

We wouldn't lose any practical modeling power by dropping power-exponential and supporting only Matérn, and Matérn allows for a much broader range of smoothness classes, so unless there's a very particular reason to retain the power-exponential kernel, in the interest of simplicity we should drop it.

maresb:
This PR replaces the power-exponential kernel with the more flexible Matérn kernel.

#### Replace power-exponential with Matérn kernel
- Use `sklearn.gaussian_process.kernels.Matern` instead of custom power-exponential implementation
- The Matérn kernel allows a broader range of smoothness classes (ν parameter) and avoids positive-definiteness issues that power-exponential has for exponents > 2
- With default parameters (`matern_nu=0.5`, `matern_length_scale=1.0` diameters), the new implementation produces identical results to the old one (verified at machine precision)

#### Speed up eigenvalue computation
- Switch from `scipy.linalg.eigh` to `scipy.sparse.linalg.eigsh` (ARPACK)
- Benchmarks on the wna mesh (N=1841 triangles) show **2-6x speedup** for k=10-50 eigenvalues
- Both methods produce identical results at machine precision

#### Code cleanup
- Simplify diameter calculation using `pdist` instead of `cdist`
- Remove unnecessary `np.real()` calls (symmetric matrices always have real eigenvalues)
- Use explicit variable names (`eigenvalues_ascending`, `eigenvalues_descending`) for clarity

### Testing
- Before/after comparison confirms this is a pure refactor with differences only at machine precision (~10⁻¹³)

Closes #362

### DIFF
--- a/celeri/mesh.py
+++ b/celeri/mesh.py
@@ -84,6 +84,18 @@ class MeshConfig(BaseModel):
         Tuple containing the constrained upper and lower bounds for the elastic rates on the mesh for strike-slip.
     elastic_constraints_ds : ScalarBound
         Tuple containing the constrained upper and lower bounds for the elastic rates on the mesh for dip-slip.
+    matern_nu : float
+        Matérn kernel smoothness parameter (default 1/2). Common values: 1/2 (exponential),
+        3/2 (once-differentiable), 5/2 (twice-differentiable).
+    matern_length_scale : float
+        Matérn kernel length scale (default 1.0). Interpretation depends on matern_length_units.
+    matern_length_units : Literal["absolute", "diameters"]
+        Units for matern_length_scale: 'diameters' scales by mesh diameter (default),
+        'absolute' uses the value directly in the same units as mesh coordinates.
+    eigenvector_algorithm : Literal["eigh", "eigsh"]
+        Algorithm for eigendecomposition (default "eigh"). 'eigh' (dense LAPACK) is faster for
+        many modes, 'eigsh' (sparse ARPACK) is faster for few modes. Both have equivalent accuracy,
+        but eigenvector signs may differ between algorithms.
     """
 
     # Forbid extra fields when reading from JSON
@@ -151,6 +163,12 @@ class MeshConfig(BaseModel):
     iterative_coupling_linear_slip_rate_reduction_factor: float = 0.025
     iterative_coupling_smoothing_length_scale: float | None = None
     iterative_coupling_kinematic_slip_regularization_scale: float = 1.0
+
+    # GP kernel hyperparameters for eigenmode computation
+    matern_nu: float = 0.5
+    matern_length_scale: float = 1.0
+    matern_length_units: Literal["absolute", "diameters"] = "diameters"
+    eigenvector_algorithm: Literal["eigh", "eigsh"] = "eigh"
 
     @classmethod
     def from_file(cls, filename: str | Path) -> list[MeshConfig]:
@@ -468,7 +486,7 @@ def _get_eigenvalues_and_eigenvectors(
             subset_by_index=[n_tde - n_eigenvalues, n_tde - 1],
         )
     elif eigenvector_algorithm == "eigsh":
-        # ARPACK is faster for small k, but eigenvector signs may differ
+        # ARPACK is faster for small k; eigenvector signs may differ from eigh
         eigenvalues_ascending, eigenvectors_ascending = scipy.sparse.linalg.eigsh(
             covariance_matrix, k=n_eigenvalues, which="LM"
         )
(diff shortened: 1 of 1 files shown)


celeri/mesh.py
```diff
@@ -426,41 +428,43 @@ def _compute_mesh_perimeter(mesh: dict):
...
+    # Use sklearn's Matern kernel which handles all special cases (nu=0.5, 1.5, 2.5)
+    # and the general case with proper numerical stability
+    kernel = Matern(nu=matern_nu, length_scale=matern_length_scale)
+    correlation_matrix = matern_sigma**2 * kernel(centroid_coordinates)
 
     # https://stackoverflow.com/questions/12167654/fastest-way-to-compute-k-largest-eigenvalues-and-corresponding-eigenvectors-with
-    eigenvalues, eigenvectors = scipy.linalg.eigh(
-        correlation_matrix,
-        subset_by_index=[n_tde - n_eigenvalues, n_tde - 1],
+    # Use ARPACK (eigsh) for faster computation of top k eigenvalues
```
brendanjmeade: Fascinating this this is faster.  Out of curiosity, is it because it only computes k eigenvalues vs the eigh approach, where I was calculating all eigenvalues and keeping just a few?

maresb:
> One little thing: Do we need the matern_length_units setting?

We need it in order to preserve the current behavior, because the current Matérn length scale on each mesh is 1 mesh diameter. Thus I'm currently expressing that in my defaults:

```python
matern_length_scale=1
matern_length_units="diameter"
```

I'm very interested in experimenting with a fixed length scale, and that PR enables that experiment by adding `matern_length_units="absolute"`.

maresb:
@brendanjmeade, I think this is ready to merge.

celeri/mesh.py
brendanjmeade: Facinating.  No need to dive deeper.  I'll take the performance win!

brendanjmeade:
> Is this something that could cause confusion, for example when viewing previous results? Is the eigenfunction computation even a bottleneck relative to the TDE stuff? Maybe we should drop that change, or convert it into a comment.

I'm fine with this as long as we understand it today and are consistent moving forward.

PR #368, opened 2026-01-14
add coupling, segment slip plots to plot_estimation_summary

elc45:
also drop unnecessary model argument to `plot_estimation_summary`

### DIFF
--- a/celeri/plot.py
+++ b/celeri/plot.py
@@ -595,11 +595,6 @@ def plot_estimation_summary(
             plt.title("Strike-Slip Coupling")
             common_plot_elements(segment, lon_range, lat_range)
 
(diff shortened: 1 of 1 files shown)


aseyboldt:
Thanks! Can you show an example output here in the PR?

elc45:
For that run the coupling constraints were [-5, 5] so I think we're good! But yeah normally it's set to [0,1]

brendanjmeade:
@elc45 Are you confident that this is behaving correctly with [-1, 1] coupling bounds?

elc45:
Yeah, that's right. With constraint sigma's at 1, which could be increased or decreased.

PR #370, opened 2026-01-15
Include version/commit information in `Estimation` / `remaining.json`

maresb:
I wanted to do a little something in favor of reproducibility against backwards-compatibility issues. (See https://github.com/brendanjmeade/celeri/pull/367#issuecomment-3750912310.)

If we're accepting that we may occasionally be making breaking changes between runs, then I really want that we can at least see which commit we were on when a previous run was produced.

After this PR, we get a new key in `remaining.json` that looks like:

```json
"celeri_version": "0.0.4.post1.dev289+g7182bb35f.d20260115"
```

The string here is generated `setuptools-scm` via `hatch-vcs`, and contains quite a bit of info:

`0.0.4`: the latest version was v0.0.4.
`.post1`: we're *after* the v0.0.4 release.
`.dev289`: we're 289 commits after the v0.0.4 release.
`+g7182bb35f`: we're on the **G**it commit `7182bb35f`.
`.d20260115`: the Git working area is dirty (has uncommitted changes), and the date is 2026-01-15.

In order to future-proof against situations where users install this in a non-development context (e.g. on a cluster), this extended info that relies on Git is opt-in via the `CELERI_HATCH_VCS_RUNTIME_VERSION = "1"` envvar, which pixi automatically sets for all our dev environments.

Users installing from PyPI, or `pixi` users with a clean checkout from a release version will just see something like:

```json
"celeri_version": "0.0.4"
```

### DIFF
--- a/celeri/solve.py
+++ b/celeri/solve.py
@@ -32,6 +32,7 @@ from celeri.operators import (
     rotation_vectors_to_euler_poles,
 )
 from celeri.output import dataclass_from_disk, dataclass_to_disk, write_output
+from celeri.version import __version__ as celeri_version
 
 
 @dataclass(kw_only=True)
@@ -57,6 +58,8 @@ class Estimation:
     trace: Any | None = None
     # MCMC trace from Bayesian inference.
     mcmc_trace: Any | None = None
+    # Version of celeri used to create this estimation.
+    celeri_version: str | None = None
 
     @property
     def model(self) -> Model:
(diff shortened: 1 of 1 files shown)


maresb:
Side note: the `pixi.lock` files are becoming quite large (almost 1MB), and they change often, so it's not ideal to have them in the repo like this, because Git stores the entire history, so cloning will eventually require significant bandwidth and disk space. It's not an immediate crisis, but something to consider for the long-term.

(A potential solution is Git LFS, but that has drawbacks: it requires special commands to properly clone the repo.)

aseyboldt:
I like this, that will make it easier to see if we load old runs with a new library version. Maybe we could even add a warning if you load an old run with a more recent celeri version? It *should* still work, but if something looks strange it is easier to debug if you already have a hint that that might be related...

About the pixi size, maybe we could just drop the osx-64 and win platforms? That should cut the size it half. But shouldn't most of the updates of the pixi lock file be relatively small, at least much smaller than the total file size?

PR #375, opened 2026-01-17
Fix non-deterministic smoothing matrix due to in-place input mutation

maresb:
Working on ensuring reproducible results, I noticed the following issue:

`get_tri_smoothing_matrix` mutated its `tri_shared_sides_distances` input array in-place:

```python
# Replace zero distances with 1 to avoid divide by zero
tri_shared_sides_distances[np.where(tri_shared_sides_distances == 0)] = 1
```

This caused inconsistent results depending on whether the mesh data was:
- **Fresh** (zeros in missing neighbor slots) → correct smoothing matrix
- **Previously serialized** (zeros already converted to 1s) → different smoothing matrix

The bug manifested as ~0.05% relative differences in smoothing matrix values for edge triangles, because the `share_distances` sum was computed *before* the mutation, but inverse distances were computed *after*.

### The Fix

1. Use **NaN** instead of 0 as the sentinel for missing neighbors
2. Use `np.nansum()` and let `1.0/NaN = NaN` propagate harmlessly  
3. **No input mutation** — the function is now pure
4. Vectorized implementation using COO format + Kronecker product

### Backward Compatibility

- Old estimations load correctly (smoothing matrices stored in `.npz` files)
- `Mesh.from_disk` already recomputes `tri_shared_sides_distances`, so old zarr caches are unaffected
- New runs will produce consistent, correct results

### DIFF
--- a/celeri/spatial.py
+++ b/celeri/spatial.py
@@ -1,4 +1,3 @@
-import copy
 import warnings
 
 import cutde.halfspace as cutde_halfspace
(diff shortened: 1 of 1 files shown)


maresb:
Awesome, thanks for the quick review!

PR #376, opened 2026-01-18
Add on-demand operator regeneration to reduce disk usage

maresb:
Doing lots of runs is killing my available disk space, so I threw together this PR to use the operator cache rather than duplicating them on every run.

DISCLOSURE: This was a bit vibe-coded. It should be treated with suspicion.

### Summary

Adds the ability to save estimations without storing the full operator arrays (~6.8GB per run), regenerating them from the elastic operator cache on load. This significantly reduces disk usage for multiple runs that share the same geometry.

### Changes

**Core Feature:**
- `Operators.to_disk(save_arrays=False)` saves only model and index, with a marker file for regeneration
- `Operators.from_disk()` detects the marker and regenerates operators from the elastic cache
- `Estimation.to_disk(save_operators=False)` exposes this at the estimation level

**Configuration:**
- New `save_operators` config option (default: `True` for backward compatibility)
- New `--save_operators` CLI flag

**Bug Fix:**
- Fixed `tde=False` mode in `build_operators` and `build_estimation` which was broken:
  - `_store_elastic_operators` now receives the `tde` parameter
  - Index selection uses `_get_index_no_meshes` when `tde=False`
  - `build_estimation` uses correct `_get_data_vector_no_meshes`

### Usage

```bash
# Via CLI
celeri-solve config.json --save_operators 0

# Via config file
{
  "save_operators": false,
  "elastic_operator_cache_dir": "/path/to/cache"
}
```

**Note:** Requires `elastic_operator_cache_dir` to be set for regeneration to work.

### Testing

- Added `test_operators_serialization_minimal` - tests `Operators` round-trip with regeneration
- Added `test_estimation_serialization_minimal` - tests `Estimation` round-trip with regeneration

### DIFF
--- a/tests/test_serialize.py
+++ b/tests/test_serialize.py
@@ -271,3 +271,122 @@ def test_estimation_serialization(config_file, temp_dir):
                 original_estimation.tde_rates[mesh_idx],
                 deserialized_estimation.tde_rates[mesh_idx],
             )
+
+
+@pytest.mark.parametrize(
+    "config_file",
+    [
+        "./tests/configs/test_japan_config.json",
+    ],
+)
+def test_operators_serialization_minimal(config_file, temp_dir):
+    """Test serializing operators with save_arrays=False and loading from cache.
+    
+    Uses tde=False to skip expensive TDE computation - we're testing serialization,
+    not operator correctness.
+    """
+    # Load model from config
+    config = celeri.get_config(config_file)
+
+    # Use isolated temp directory for cache to ensure test doesn't depend on
+    # pre-existing cache state and doesn't pollute shared cache
+    cache_dir = Path(temp_dir) / "operator_cache"
+    cache_dir.mkdir(exist_ok=True)
+    config.elastic_operator_cache_dir = cache_dir
+
+    model = celeri.build_model(config)
+
+    # Build operators without TDE/eigen - much faster, still tests serialization
+    original_operators = celeri.build_operators(model, eigen=False, tde=False)
+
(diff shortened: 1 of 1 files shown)


celeri/operators.py
```diff
@@ -493,18 +493,44 @@ def weighting_vector(self) -> np.ndarray:
...
-    def to_disk(self, output_dir: str | Path):
-        """Save TDE operators to disk."""
+    def to_disk(self, output_dir: str | Path, *, save_arrays: bool = True):
+        """Save operators to disk.
+
+        Args:
+            output_dir: Directory to save operators to.
+            save_arrays: If True (default), save all operator arrays to disk.
+                If False, only save model and index. On load, operators will be
+                regenerated using the elastic operator cache, saving ~6.8GB per run.
```
aseyboldt: same

maresb: Fixed in f9867a1

celeri/operators.py
```diff
@@ -650,7 +703,7 @@ def build_operators(model: Model, *, eigen: bool = True, tde: bool = True) -> Op
     operators = _OperatorBuilder(model)
 
     # Get all elastic operators for segments and TDEs
-    _store_elastic_operators(model, operators)
+    _store_elastic_operators(model, operators, tde=tde)
```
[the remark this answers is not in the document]
maresb: Ya, it just wasn't being passed through before f6da2c0. It would have been flagged by the linter as unused if it weren't for the usage in the `ValueError` just before.

maresb:
Rebased to squash review and improve consistency

PR #393, opened 2026-01-26
Deprecate and remove dead snap_segments code

maresb:
## Summary

- Add deprecation warning when `snap_segments` is explicitly set in config files
- Raise error if `snap_segments` is set to a truthy value (feature removed)
- Remove dead `snap_segments` and `make_default_segment` functions from `model.py` (133 lines)
- Remove `snap_segments` from all tracked config files

## Background

The `snap_segments` config option was:
- Always set to 0 in every config file (never enabled)
- Broken due to deprecated `DataFrame.append()` removed in pandas 2.0
- Duplicated from `celeri/scripts/snap_segments.py` which is the working, tested version

Users who need segment snapping should use the standalone script instead.

### DIFF
--- a/data/config/japan_config.json
+++ b/data/config/japan_config.json
@@ -24,7 +24,6 @@
   "sar_file_name": null,
   "segment_file_name": "../segment/japan_segment.csv",
   "slip_constraint_weight": 100000,
-  "snap_segments": 0,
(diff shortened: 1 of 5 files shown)


brendanjmeade:
@jploveless Any words about this deprecation of snap segments?

PR #420, opened 2026-02-19
Add --description and --run-dir-file CLI args + eigenvalues backward compat

maresb:
## Summary

Split from #414.

- Add `--description` and `--run-dir-file` CLI arguments to `celeri-solve`
- Add `Config.description` field for tagging runs
- Make `EigenOperators.eigenvalues` backward-compatible with older saved models via `field(default_factory=dict)`

### DIFF
--- a/celeri/cli.py
+++ b/celeri/cli.py
@@ -254,6 +254,21 @@ def parse_args() -> argparse.Namespace:
         help="Default Matérn length scale for mesh eigenfunctions",
     )
 
+    parser.add_argument(
+        "--description",
+        type=str,
+        default=None,
+        required=False,
+        help="Human-readable description or tag for this run",
+    )
+    parser.add_argument(
+        "--run-dir-file",
+        type=str,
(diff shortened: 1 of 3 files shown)


maresb:
Context behind `--run-dir-file`: when I run a parameter scan, I spawn a bunch of CLI `celeri-solve` processes. Each process creates a new run directory. Then after-the-fact, I must somehow match up the parameters and the runs. This is challenging because I have to index all the possible run directories, make a key with the varied scan parameters, and do a lookup. It's also not robust in case I somehow accidentally run `celeri-solve` by hand in the meantime. So this implements a very crude form of IPC (inter-process communication) by writing the name of the run directory to a file.

Thus in my parameter scan script, I can do `celeri-solve --run-dir-file /tmp/some-unique-filename.txt ...`. Then I can read `/tmp/some-unique-filename.txt` and read out the path as `runs/00000000123`, skipping the whole run directory detection drama.

PR #424, opened 2026-02-19
Allow Unicode math symbols in docstrings and comments

maresb:
## Summary

Scientific code naturally uses σ, ×, →, π etc. in documentation. This PR disables the Ruff rules RUF001/RUF002/RUF003 globally so that Unicode math symbols are allowed in strings, docstrings, and comments.

This unblocks upcoming documentation PRs that add unit annotations with Unicode symbols.

## Changes

- Add RUF001, RUF002, RUF003 to the global `ignore` list in `pyproject.toml`
- Update `pixi.lock`


maresb:
Ya, in product-oriented software it's a security risk because there are all sorts of screwy tricks one can pull with Unicode, but for scientific software it makes so much more sense to allow it.

PR #427, opened 2026-02-19
Enable relative path serialization for Config (resolves #418)

maresb:
## Summary

- Extract the `paths_relative_to` serialization logic from `MeshConfig` into a reusable `RelativePathSerializerMixin` in `celeri_util.py`
- Add the mixin to both `MeshConfig` and `Config`, so that `segmesh.py` (and other scripts) can write config files with relative paths
- Includes @jploveless's `segmesh.py` changes from #418, merged with current `main`

This resolves the issue in #418 where `model.config.model_dump(mode="json", context={"paths_relative_to": ...})` silently ignored the context because `Config` lacked the corresponding `model_serializer`.

## Test plan

- [x] Existing `test_mesh_config_relative_path_context` passes
- [x] Verified `Config.model_dump(mode="json", context={"paths_relative_to": ...})` produces relative paths
- [x] Merge conflict with `main` resolved (only in `mesh.py` class declaration)

### DIFF
--- a/celeri/cli.py
+++ b/celeri/cli.py
@@ -253,6 +253,57 @@ def parse_args() -> argparse.Namespace:
         required=False,
         help="Default Matérn length scale for mesh eigenfunctions",
     )
+    parser.add_argument(
+        "--mcmc-default-mesh-gp-parameterization",
+        type=str,
+        default=None,
+        required=False,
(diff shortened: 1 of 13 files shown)


maresb:
@jploveless, it's fine from our side for you to merge this. It won't conflict with any of our current work.

PR #429, opened 2026-02-20
Add examples for plotting posterior stdevs

elc45:
Adding to `doc_matplotlib_vis.ipynb` examples for how to see a) standard deviations of slip/coupling rates on a mesh (_given_ a draw or aggregation of draws), and b) MCMC standard deviations _between_ draws, on the tde-level.


aseyboldt:
Sorry, saw it only just now:
It says "Average standard deviation", but it's the minimum standard deviation.

Issue #436, opened 2026-03-11
BEL variance captured


brendanjmeade:
When running with the new mesh truncation error checker and the latest [WNA model](https://github.com/brendanjmeade/wna/pull/30) (with BEL `matern_length_scale = 0.025`, and 80 modes updated per discord)

I'm seeing:

```
INFO       Rough min length scale for 80 modes (1⌀ = 1245728m): ℓ(90.0%)≈0.03⌀, ℓ(95.0%)≈0.03⌀,
           ℓ(99.0%)≈0.05⌀, ℓ(99.9%)≈0.08⌀
           - celeri.mesh:_log_eigenvalue_truncation:919 - 2026-03-10 22:03:47
WARNING    Only 74.42% of prior variance captured by 80 eigenmodes. Consider increasing
           n_eigenvalues or matern_length_scale to reduce truncation artifacts.
```

Should I push `matern_length_scale` higher?

maresb:
Oops, in my last message, I meant to write "Yes, you should set `l=0.06`." This is what's now recommended by the log messages to reach 95% after my fix, and it works. Sorry to give you a general answer while forgetting to include the specific guidance.

Issue #374, opened 2026-01-15
Understanding truncation error due to `n_eigenvalues`


maresb:
I teased this topic in the last meeting.

The more eigenmodes we keep, the better our GP approximation becomes. I'll answer the question of how the results change as the number of retained eigenmodes `n_eigenvalues` increases, at least as far as the prior is concerned.

### Notation

* $N$: number of retained eigenmodes (`n_eigenvalues`)
* $\nu$: Matérn smoothness parameter
* $\ell$: Matérn length scale
* $d$: dimensionality of the mesh, in our case $d=2$
* $A$: area of the mesh (our current area measure is that each triangle contributes 1 unit, so until we change it, this is the integer number of triangles)

### Claim

In the manifold limit (the mesh is made arbitrarily fine), the relative truncation error (fraction of prior variance discarded) for the prior GP is asymptotic as $N\to\infty$ to

$$
C(d,\nu) \left( \frac{\ell^d N}{A} \right)^{-2\nu/d}
$$

for some explicit constant $C(d,\nu)$ independent of $N$, $\ell$, and $A$.

In the two-dimensional case $d=2$, relative error is asymptotic to

$$
\left(\frac{2\pi\ell^2 N}{\nu A}\right)^{-\nu}.
$$

Suppose we fix a mesh (fixes $A$), and an error tolerance (say $\epsilon=10\\%$), then we can solve for $N$:

$$
N \approx \left(\frac{\nu A}{2\pi\ell^2}\right)\epsilon^{-1/\nu}.
$$

As expected $N$ increases when you decrease $\ell$. Interestingly, $N$ decreases as the smoothness $\nu$ increases, at least focusing on the $\epsilon$ factor: this makes sense because the smoother the prior, the less important the higher modes are.

### Some more details

The $i$-th eigenmode independently contributes $\lambda_i$ to the variance, so the total variance of the GP is $\sum \lambda_i$. Truncating the number of eigenmodes discards some portion of that total variance.

The full formula is

$$
\frac{\sum_{n>N}\lambda_n}{\sum_{n\ge1}\lambda_n}
\sim
\frac{1}{\nu\\,B(\nu,d/2)}
\left(\Gamma\left(\frac d2+1\right)\frac{(\sqrt{2\pi/\nu}\\,\ell)^d\\,N}{A}\right)^{-2\nu/d}.
$$

[...]

PR #437, opened 2026-03-11
Fix Weyl length scale estimates

maresb:
## Summary

- Format ℓ estimates with 2 significant figures (`.2g`) so that e.g. `ℓ(99.0%)≈0.024⌀` instead of `ℓ(99.0%)≈0.02⌀`.
- Cap the Weyl correction factor at 1.0 for ℓ estimates. When the mesh is underprovisioned (C < 1), the correction was calibrated far from the Weyl asymptotic regime and made ℓ estimates dramatically worse. With the cap, the raw Weyl estimate is used instead, which is conservative but much more reliable.

Addresses #436 — the ℓ estimates shown to @brendanjmeade were distorted by the uncapped correction. With this fix, the estimates are more useful for deciding how to adjust `matern_length_scale`.

Refs: #374

### DIFF
--- a/celeri/mesh.py
+++ b/celeri/mesh.py
@@ -904,10 +904,14 @@ def _log_eigenvalue_truncation(
             and diameter is not None
             and matern_length_units is not None
         )
+        # Cap the correction at 1.0 for ℓ estimates: when the mesh is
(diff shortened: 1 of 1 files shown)


maresb:
This PR only affects logging, so merging.

PR #455, opened 2026-05-02
feat: configurable sigmoid slope for two-sided coupling constraint

brendanjmeade:
The two-sided sigmoid in `_constrain_field` previously hard-coded a slope factor of 4 (giving unit slope at the midpoint), which causes modest excursions of the latent GP to saturate near the bounds and tends to produce near-binary coupling fields. Expose this as `mcmc_default_mesh_sigmoid_slope` (default 4.0, preserves existing behavior) on `Config`, mirror as `MeshConfig.sigmoid_slope` for per-mesh overrides, and thread through `_constrain_field`, `_unconstrain_field`, and `_get_unconstrained_mean`. Smaller values widen the transition zone and allow more intermediate coupling values.

### DIFF
--- a/celeri/config.py
+++ b/celeri/config.py
@@ -340,6 +340,25 @@ class Config(RelativePathSerializerMixin, BaseModel):
     in the per-mesh configuration.
     """
 
+    mcmc_default_mesh_sigmoid_slope: float = 4.0
+    """Default slope factor for the two-sided sigmoid constraint transform.
+
+    Controls how sharply the unconstrained latent field is squashed into the
+    bounded interval ``[lower, upper]`` (e.g. coupling in [0, 1]). The
+    transform is ``sigmoid(slope * (x - midpoint) / scale) * scale + lower``,
+    so the derivative at the midpoint equals ``slope / 4``.
+
+    - ``4.0`` (default) gives unit slope at the midpoint, preserving the
+      historical behavior — modest excursions saturate near the bounds,
+      tending to produce near-binary coupling fields.
+    - Smaller values (e.g. ``1.0``) widen the transition zone in latent
+      space, allowing more intermediate constrained values.
+    - Larger values produce sharper, more step-like transitions.
+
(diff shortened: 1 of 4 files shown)


maresb:
@brendanjmeade I'm a bit skeptical this knob is adding expressive power. Recalling our **brightness/contrast** framing — `coupling_mean` is brightness, `coupling_sigma` is contrast — `sigmoid_slope` looks like a second contrast knob that's redundant with `coupling_sigma`.

In `_constrain_field`, the latent `f` only enters the sigmoid as `(slope/scale) · (f − midpoint)`, and `f − μ_unc` has standard deviation proportional to `coupling_sigma`. So the prior on the constrained field depends on `slope` and `coupling_sigma` only through their product (i.e. only through the combined contrast):

> `(coupling_sigma=σ, sigmoid_slope=s)` ≡ `(coupling_sigma=σ·s/4, sigmoid_slope=4)`

On the brightness side: with the default `coupling_mean_parameterization="constrained"`, `_get_unconstrained_mean` divides by `slope`, which exactly cancels the `slope` factor inside the sigmoid — so brightness is unaffected.

So the "smaller slope → more intermediate coupling values" effect described in the docstring is the same thing as "lower contrast via smaller `coupling_sigma`" — just redundant config rather than a new degree of freedom.

brendanjmeade:
Ahh.  Thanks @maresb for such a clear explanation!  Leave it to me to be wrong or, at best, redundant.

So the more appropriate tunable knob is:

For default brightness/contrast (more binary coupling):
`"mcmc_default_mesh_coupling_sigma": 1.0,`

For less brightness/contrast (less binary coupling):
`"mcmc_default_mesh_coupling_sigma": 3.0,` <--- should this increase or decrease?

maresb:
`coupling_sigma` is the GP amplitude that's input to the sigmoid function. If you want less contrast/binary then you should make the `coupling_sigma` value smaller.

PR #460, opened 2026-05-21
Process horizontal meshes

jploveless:
For truly horizontal meshes, which can be easier to use compared to slightly dipping meshes for BELs that share one side with the deep edge of a subduction zone, `mesh.py` would fail because of some of the depth checking that is done to assign edge-lining elements to top, bottom, and side. This minor change to `mesh.py` just prevents two (very similar) conditional statements from being entered when doing some refinement of adding elements to the `top` and `bottom` lists for horizontal meshes. Those conditional statements were based on histograms of depths along the top and bottom of a mesh, which is not meaningful for horizontal meshes. The change just says that a histogram 

The net result of this change is that all edge-lining elements of a horizontal mesh are assigned to be `side` elements, so technically they could still have edge slip rate constraints applied, with that categorization in mind. 

For the 3 Japan 187 Western US meshes, no differences in the edge element categorization resulted from this change, so I think it's pretty unobtrusive.

### DIFF
--- a/celeri/mesh.py
+++ b/celeri/mesh.py
@@ -565,7 +565,7 @@ def _compute_mesh_edge_elements(mesh: dict):
     depth_bin_min = depth_bins[0:-1]
     depth_bin_max = depth_bins[1:]
     zero_idx = np.where(depth_count == 0)[0]
-    if len(zero_idx) > 1:
+    if len(zero_idx) > 0 & len(depth_count) > 1:
         if np.abs(depth_bin_max[zero_idx[0]] - depth_bin_min[zero_idx[-1] + 1]) > 10:
             tops[centroid_depths < depth_bins[zero_idx[0]]] = False
     # Assign in to meshes dict
@@ -593,7 +593,7 @@ def _compute_mesh_edge_elements(mesh: dict):
     depth_bin_min = depth_bins[0:-1]
     depth_bin_max = depth_bins[1:]
     zero_idx = np.where(depth_count == 0)[0]
-    if len(zero_idx) > 1:
+    if len(zero_idx) > 0 & len(depth_count) > 1:
         if abs(depth_bin_min[zero_idx[-1]] - depth_bin_max[zero_idx[0] - 1]) > 10:
             bots[centroid_depths > depth_bin_min[zero_idx[-1]]] = False
     # Assign in to meshes dict
--- a/celeri/mesh.py
+++ b/celeri/mesh.py
@@ -565,7 +565,7 @@ def _compute_mesh_edge_elements(mesh: dict):
     depth_bin_min = depth_bins[0:-1]
     depth_bin_max = depth_bins[1:]
     zero_idx = np.where(depth_count == 0)[0]
-    if len(zero_idx) > 0:
+    if len(zero_idx) > 1:
         if np.abs(depth_bin_max[zero_idx[0]] - depth_bin_min[zero_idx[-1] + 1]) > 10:
             tops[centroid_depths < depth_bins[zero_idx[0]]] = False
     # Assign in to meshes dict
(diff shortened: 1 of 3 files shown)


celeri/mesh.py
```diff
@@ -565,7 +565,7 @@ def find_matching_rows(array1, array2):
     depth_bin_min = depth_bins[0:-1]
     depth_bin_max = depth_bins[1:]
     zero_idx = np.where(depth_count == 0)[0]
-    if len(zero_idx) > 0:
+    if len(zero_idx) > 1:
```
brendanjmeade: @jploveless Can you explain the change from `0` to `1` here?  Is this the correcting a pre-exiting error?

brendanjmeade (commented):
Before merging, I want to understand the `len(zero_idx)` change below.  Interested!

Moving to all side boundary conditions for perfectly flat BEL meshes sounds sensible.  However, a caveat is that I don't think (maybe?) we implemented side boundary conditions for the MCMC solve.  I don't want this change to break anything that is already working with the WNA + BEL model.

celeri/mesh.py
```diff
@@ -565,7 +565,7 @@ def find_matching_rows(array1, array2):
     depth_bin_min = depth_bins[0:-1]
     depth_bin_max = depth_bins[1:]
     zero_idx = np.where(depth_count == 0)[0]
-    if len(zero_idx) > 0:
+    if (len(zero_idx) > 0) & (len(depth_count) > 1):
```
brendanjmeade, line 568: @jploveless Is it accurate to say that the `len(depth_count) > 1` is a check for non-flat (non-constant depth)?

jploveless: I wish I had documented this secondary check better! 

The assignment of edge-lining elements to top and bottom is based on the criteria: 
```
# Top elements are those where the depth difference between the non-edge node
# and the mean of the edge nodes is greater than the depth difference between
# the edge nodes themselves
```
and
```
# Bottom elements are those where the depth difference between the non-edge node
# and the mean of the edge nodes is more negative than the depth difference between
# the edge nodes themselves
```
but there are some rogue truly side nodes that can meet these criteria and can be initially assigned as top or bottom (near the SW edge of the Japan Trench, for example). 

The secondary check I introduced aims to solve this, taking the distribution of element centroid depths across all elements inititally flagged as top or bottom. Then, any elements that effectively lie within their own depth bin (a bin separated from other depth bins) is reset to be a side element. This has worked for the few meshes I've checked it with. 

You are right to be suspicious of my initial change from `0` to `1` and the most recent commit goes back to `0` but introduces an additional condition that rules out satisfying the `if` statement for perfectly flat meshes: there must be more than one depth bin, which is always(?) the case for a dipping mesh (even the very gentle WNA BEL) but importantly NOT the case for perfectly flat meshes. 

The most recent change should have no impact on current/past models, because none has incorporated a flat mesh, and I did a better check this time to make sure that the classification of edge elements is identical with this code as with what's currently on `main`. The WUS BEL is not very predictable in terms of its edge element classification, because it's so irregularly shaped in its perimeter. So, I think you've done the right thing of setting ALL edge slip constraints to 1, and that makes me think that the side constraints are properly handled throughout the run (this could be confirmed by making sure that there's zero slip estimated around the entire perimeter).

brendanjmeade: Thanks for taking the time to explain this so carefully.  I'm not certain about the histogram approach.  Is the goal just to check for different depths on a given mesh?  If so, would this have an advantage over something like counting the number of unique depths?  It feels like the hist results might be used for something more?

jploveless: The goal of the histogram is to see which elements deviate in depth substantially from true top or bottom elements. Because some side elements are initially tagged as top/bottom, I use the depth distribution of all initially tagged elements to try to separate out what should be classified as a side element. Unique depths wouldn't work because the centroid depths for true top/bottom elements will be variable, and I didn't want to just use edge depths because that won't work for meshes that aren't generated based on depth contours (i.e. the top/bottom mesh edge might be variable depth). The histogram results really are just used for un-assigning actual side elements that were initially tagged as top or bottom. Here's an example for the southern end of the Japan Trench. The uncorrected side just comments out the `if` statements, and the right side uses them. 
side_compare

I definitely think there could be meshes that don't work well with this approach. I've tried a bunch of approaches to assign labels to edge elements, and this seems the best so far, but I'm sure there are more versatile, less convoluted methods!

brendanjmeade: Great figure, and I understand this now.  It's definitely a tricky calculation, and that makes it important!  A working understanding of the algorithm and some rules of thumb will take this far.  Thanks for taking the time to explain this!

PR #465, opened 2026-06-03
Honor --mesh_parameters_file_name CLI override (issue #462)

### LINKED ISSUE #462: `--mesh_parameters_file_name` CLI not recognized
**!!!**
**This isn't urgent/necessary, because there's an easy workaround, but I wanted to flag this anyway.** 
**!!!**

I'm trying to test the effect of various mesh parameters (number of modes, etc.) and so I've created a few alternate mesh parameter files. I call `celeri-solve` with the command line option of `--mesh_parameters_file_name mesh/japan_mesh_alt0.json`. In the log, I see the warning: 

```
WARNING    ORIGINAL: config.mesh_parameters_file_name: /Users/jloveles/Documents/python/japan_imaging/celeri_inputs/mesh/japan_mesh.json             
           - celeri.cli:process_args:327 - 2026-06-02 20:42:43                                                                                       
WARNING    REPLACED: config.mesh_parameters_file_name: /Users/jloveles/Documents/python/japan_imaging/celeri_inputs/mesh/japan_mesh_alt0.json        
           - celeri.cli:process_args:329 - 2026-06-02 20:42:43  
```

but the processed mesh parameters reflect those in the mesh parameter file referenced in the `.config` file. Printing the entire `model.mesh_config` object shows that the originally referenced file is being read, as its filename is shown with each mesh. I've tried specifying absolute paths for the alternate file, but that doesn't work either. 

Changing other input files (such as the station file) seems to be done correctly: I see the warning about the change and then the `SUCCESS` message in the log, indicating that the command line option was recognized.

In `main` of `celeri_solve.py`, the [CLI args are processed, and THEN `build_model` is called](https://github.com/brendanjmeade/celeri/blob/2c80b67e35343a3772889dc15edf7102be5cd13c/celeri/scripts/celeri_solve.py#L23), and I think that part of this calls `get_config`, which [contains the line](https://github.com/brendanjmeade/celeri/blob/2c80b67e35343a3772889dc15edf7102be5cd13c/celeri/config.py#L658): 

```
mesh_parameters_file_name = config_data.get("mesh_parameters_file_name", None)
```
So I think that, even if `--mesh_parameters_file_name` is changed on the command line, that change is ignored because the variable is defined from `config_data`, which is read directly from the `config.json`. 

I'm scared to tinker too much because I know there's a lot of the propagation of default mesh parameters that happens, and I don't want to screw that up. But it seems like the parsed command line arguments need to get passed into `get_config` so that a change to the mesh parameter file name can be recognized. 

For now, I'm just going to create a `.config.json` file for each of the alternate mesh parameter files I want to test, which is very straightforward.

brendanjmeade:
## Summary

Fixes #462. The `--mesh_parameters_file_name` command-line override to `celeri-solve` was logged as applied but silently ignored.

**Root cause:** `Config` holds both `mesh_parameters_file_name` (the path) and `mesh_params` (the eagerly-loaded per-mesh parameter list). `get_config` loads `mesh_params` from the original config file, and `build_model` consumes `mesh_params` — but `process_args` only updated the *path* field. So the override produced the `ORIGINAL`/`REPLACED` log lines without changing the parameters actually used. (Station/segment/block overrides work because those files are read by path at build time, not pre-loaded.)

## Changes

- **`celeri/config.py`** — Extract the path→`mesh_params` load into a shared `load_mesh_params(path, base_dir)` helper; `get_config` now calls it (pure refactor, behavior unchanged).
- **`celeri/cli.py`** — `process_args` re-loads `config.mesh_params` via the helper whenever `--mesh_parameters_file_name` is supplied. CLI paths resolve relative to the cwd, matching the other `--*_file_name` overrides, and a `SUCCESS Read:` line is logged. `propagate_mesh_defaults` still runs later in `read_data`, so inherited mesh defaults are filled from the (already-overridden) top-level config.
- **`tests/test_process_args.py`** — New regression tests for the override and the no-override case (uses `tmp_path`, no committed fixture).

## Why this approach

A fuller redesign making `mesh_params` lazily derived from the path would break the load→mutate→build pattern that `test_optimize.py` and `scripts/segmesh.py` depend on — segmesh points the path at a file it *writes from* `mesh_params`, which makes a path→list derivation circular. This localized fix removes the path/list drift without disturbing those flows.

## Note

A CLI-supplied **relative** mesh-params path now resolves relative to the **cwd** (consistent with the other file-name overrides), which differs from the config-dir-relative resolution used for the path inside the config JSON. Absolute paths are unambiguous either way.

## Testing

- New regression tests pass; reproduction confirms `mesh_params` now reflects the alternate file.
- `tests/test_serialize.py` (exercises the refactored `get_config`/`MeshConfig.from_file` round-trip) passes — 13 passed total.
- `ruff check` and `ruff format --check` clean.

### DIFF
--- a/celeri/cli.py
+++ b/celeri/cli.py
@@ -1,8 +1,9 @@
 import argparse
+from pathlib import Path
 
 from loguru import logger
 
-from celeri.config import Config
+from celeri.config import Config, load_mesh_params
 
 
 def str2bool(v):
@@ -327,3 +328,16 @@ def process_args(config: Config, args: argparse.Namespace):
                     logger.warning(f"ORIGINAL: config.{key}: {original_val}")
                     setattr(config, key, args_val)
                     logger.warning(f"REPLACED: config.{key}: {args_val}")
+
+    # The per-mesh parameter list (config.mesh_params) is loaded eagerly in
+    # get_config from the original config file, so overriding only the path
+    # above would be silently ignored by build_model (issue #462). Re-load the
+    # list whenever --mesh_parameters_file_name is supplied. CLI paths are
+    # resolved relative to the cwd, matching how other --*_file_name overrides
+    # are consumed (e.g. station/segment/block files are read by path at build
+    # time). propagate_mesh_defaults still runs later in read_data, so inherited
+    # defaults are filled in from the (already overridden) top-level config.
+    mesh_params_path = getattr(args, "mesh_parameters_file_name", None)
+    if mesh_params_path is not None:
+        config.mesh_params = load_mesh_params(Path(mesh_params_path), Path.cwd())
+        logger.success(f"Read: {mesh_params_path}")
--- a/celeri/config.py
+++ b/celeri/config.py
@@ -636,6 +636,27 @@ def _get_output_path(base: Path) -> Path:
         )
 
 
+def load_mesh_params(
+    mesh_parameters_file_name: Path | str | None,
+    base_dir: Path,
+) -> list[MeshConfig]:
+    """Load the per-mesh parameter list from a mesh-parameters JSON file.
+
+    Relative paths are resolved against ``base_dir`` (the config-file directory
+    when called from ``get_config``, or the current working directory when
+    re-loading after a command-line override in ``process_args``). Returns an
+    empty list when ``mesh_parameters_file_name`` is None.
+
+    Note: the returned configs still have ``None`` for every field that inherits
+    from the top-level Config; those are filled in later by
+    ``Config.propagate_mesh_defaults`` (deferred until ``read_data`` so that CLI
+    overrides take effect first).
+    """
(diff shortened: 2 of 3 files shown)


brendanjmeade:
@jploveless It was a small bit of wrestling with Claude, but I think we got there.

jploveless:
Running this with the command I had been using correctly loads the alternate mesh file. The `SUCCESS` message is a good reassurance. I think I follow the logic laid out in the explanation and the code changes, and it seems that it really boils down to an ordering issue. Loading the mesh params when doing the CLI arg parsing seems to be the trick, as this is done after the original loading of the config file. 

Thanks for fixing this so quickly!

brendanjmeade:
@jploveless Thanks for finding this and so clearly explaining this!  This was a Claude fix with a little pushing.

PR #466, opened 2026-06-11
Make okada_wrapper an optional dependency (closes #458)

### LINKED ISSUE #458: Remove `okada_wrapper` dependency?
I'm unfortunately trying to install `celeri` on a Windows 11 machine, and I've been having a rough time. `pip`, `conda`, and `uv` installs seemed to have some version conflicts (`pydantic` was throwing a bunch of "unrecognized option" errors), and `pixi` was failing on the `okada_wrapper` install because of compiler issues. I just deleted the `okada_wrapper` dependency from `pixi.toml` and installed with `pixi`, and everything seems to be working beautifully. I know that rectangular segments are now replaced by 2 calls to `cutde` for their elastic calculations, but I'm wondering if there are any potential downstream consequences of removing the `okada_wrapper` install. If not, could it be officially removed from `pixi.toml` in the repo?

manishpaulish:
Closes #458.

This makes okada_wrapper an optional dependency rather than removing it entirely.

The package is removed from the pypi dependencies in pixi.toml, with the corresponding pixi.lock update, which fixes installs on systems where the Fortran build fails, such as the Windows 11 case described in this issue.

The triangulation="okada" runtime option in cutde_okada.py is preserved. Its lazy import now raises an informative ImportError with install instructions instead of a bare ModuleNotFoundError.

Rather than deleting the okada validation tests, the okada parametrized cases in tests/okada/test_okada.py now skip when the package is absent and still run wherever it is installed, so the cross validation against the analytical solution is retained. tests/test_okada_equals_cutde.py uses pytest.importorskip so collection no longer errors.

Validation: with okada_wrapper absent, the okada test files give 300 passed and 64 skipped on Linux with Python 3.12, and the full suite runs via pixi on macOS arm64 with Python 3.13.13 apart from the preexisting platform failures noted below.

Additional install evidence: on macOS arm64 with Python 3.13, the okada_wrapper 24.6.15 build produces a wheel that then fails the compatibility check ("not compatible with the current Python 3.13 on macOS aarch64"), so the dependency currently blocks pixi installs on Apple Silicon as well as Windows.

One unrelated observation from validating: 36 preexisting test_width_combinations degenerate_strike failures occur on macOS arm64 with all cutde triangulations, while the same cases pass on Linux on this same branch. Happy to open a separate issue with tracebacks if useful.

Also happy to adjust if you would prefer to remove the okada code path entirely.


brendanjmeade:
@manishpaulish Thank you so much for this.  This appears thoughtful and clean.  I'll merge unless there are any objections, @jploveless / @maresb?  One less dependency can only be a good thing!

manishpaulish:
Thank you, glad it could help! I will open a separate issue shortly for the macOS arm64 degenerate_strike test failures with full tracebacks, since those are unrelated to this change but probably worth tracking.

maresb (approved):
Looks excellent, thanks @manishpaulish !!!

Thanks also for the cutde issue. Am afk atm but will have a look soon

maresb:
This should be safe to merge once the conflict is resolved.

brendanjmeade:
@manishpaulish The merge conflict is due to a stale pixi.lock file. To fix it, run:

```
git checkout okada-wrapper-optional
pixi lock
git add pixi.lock
git commit -m "Regenerate pixi.lock after removing okada_wrapper"
git push
```

maresb:
@brendanjmeade, I've resolved the conflict. I think right now is the right time to get this in.

PR #469, opened 2026-06-26
Fix celeri-reweight-stations entry point (package is un-installable)

maresb:
## Problem

celeri **0.0.5 cannot be `pip install`ed at all**. The `celeri-reweight-stations` console script points at a hyphenated module path:

```toml
celeri-reweight-stations = "celeri.scripts.celeri-reweight-stations:main"
```

Python module names can't contain hyphens, and the real file is `celeri_reweight_stations.py`. pip parses the spec as module `celeri.scripts.celeri` with no callable and aborts the **entire** wheel install:

```
ERROR: For req: celeri==0.0.5. Invalid script entry point:
<ExportEntry celeri-reweight-stations = celeri.scripts.celeri:None []>
- A callable suffix is required.
```

This breaks `pip install celeri` for everyone and blocks the conda-forge build.

## Fix

Point the entry point at the real underscore module:

```diff
-celeri-reweight-stations = "celeri.scripts.celeri-reweight-stations:main"
+celeri-reweight-stations = "celeri.scripts.celeri_reweight_stations:main"
```

## Verification

Reproduced with the same step conda-forge runs (`pip install . --no-deps --no-build-isolation`): before the change the install fails with the error above; after it, `Successfully installed celeri-0.0.5`.

## Note

This needs a new release (0.0.6) for the fix to reach users/conda-forge, since 0.0.5 is already published broken on PyPI. A separate PR (#TBD) fixes the `celeri-check-segment-block-integrity` entry point, which is broken at runtime (but doesn't block install). It would also be worth having CI actually `pip install` the built wheel — `hynek/build-and-inspect-python-package` builds and inspects but never installs, which is why this slipped through to PyPI.

---
> 🤖 **AI-generated disclosure:** This PR was authored by Claude (Anthropic's Claude Code) at the request of @maresb. Please review before merging.


brendanjmeade:
This is an excellent fix.  Thanks!

PR #470, opened 2026-06-26
Move celeri_check_segment_block_integrity to notebooks/ and drop its entry point

maresb:
## Problem

`celeri/scripts/celeri_check_segment_block_integrity.py` is a VS Code-style interactive notebook (`# %%` cell delimiters), **not** a CLI. It has no `main` function and runs all of its work at **import time**, including:

```python
config_file_name = "./data/config/wna_config_constraints.json"
model = celeri.build_model(config_file_name)
```

…against a hard-coded relative path. Yet it was wired up as a console script:

```toml
celeri-check-segment-block-integrity = "celeri.scripts.celeri_check_segment_block_integrity:main"
```

So the `celeri-check-segment-block-integrity` command could never run (`ImportError: cannot import name 'main'`). (Unlike the `celeri-reweight-stations` bug in #469, this doesn't block `pip install` — it's a runtime failure.)

## Fix

Treat it as what it is — an interactive notebook:

- **Move** `celeri/scripts/celeri_check_segment_block_integrity.py` → `notebooks/celeri_check_segment_block_integrity.py` (content unchanged; `# %%` cells preserved). There it sits with the other interactive notebooks and is excluded from both the sdist and ruff linting.
- **Remove** the `celeri-check-segment-block-integrity` entry point from `pyproject.toml`.

> _Revised from an earlier draft that tried to convert this into an argparse CLI — that misread the intent. The `# %%` markers make it a notebook, so moving it is the right call._ (Branch name still says `-cli` from that first attempt.)

---
> 🤖 **AI-generated disclosure:** This PR was authored by Claude (Anthropic's Claude Code) at the request of @maresb. Please review before merging.

### DIFF
--- a/celeri/scripts/celeri_check_segment_block_integrity.py
+++ b/celeri/scripts/celeri_check_segment_block_integrity.py
@@ -1,41 +1,57 @@
-# %% Imports
+"""Plot segment/block diagnostics to check model segment and block integrity."""
+
+import argparse
+
 import matplotlib.pyplot as plt
-import matplotlib_inline.backend_inline
 import numpy as np
 
 import celeri
 
-matplotlib_inline.backend_inline.set_matplotlib_formats("retina")
 
+def main():
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "config_file",
+        help="Path to the model config JSON file (e.g. wna_config_constraints.json).",
+    )
+    parser.add_argument(
+        "--block-label",
+        type=int,
+        default=92,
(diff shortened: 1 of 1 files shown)


maresb:
Looks like this is actually a notebook, not a script, so the right resolution is probably to move it to notebooks and remove the entrypoint.

maresb:
Closed/superseded by #471 — GitHub got stuck reopening this after a force-push sync glitch. Same branch (`fix-check-segment-block-integrity-cli`) and same final commit; please review #471 instead.

PR #472, opened 2026-06-27
CI: smoke-test and functional-test the package between build and publish

maresb:
## Why

`hynek/build-and-inspect-python-package` **builds and inspects** the dists but **never installs** them, so a malformed entry point (rejected by pip only at install time) — or an installable-but-insufficient dependency set — can ship to PyPI. That's how 0.0.5 shipped un-installable (#469).

## What — two jobs between build and publish, both gating `publish-package`

1. **`smoke-test-package`** (fast, `--no-deps`): installs the built wheel and sdist, so a malformed `[project.scripts]` entry fails immediately, then statically resolves every console script against the installed sources (without importing celeri) to catch a missing module or missing `main`.

2. **`functional-test`** (real solve): installs the wheel with its **full dependency set** and runs a genuine end-to-end MCMC solve — `build_model` + `solve_mcmc(tune=10, draws=10, chains=2)` on the `wna` model with the default (jax) backend — exercising cutde, pymc, pytensor, nutpie, jax, numba, h5py, meshio and scipy together, so an installable-but-insufficient dependency set fails here. It runs from a scratch dir against a repo checkout for the example data (excluded from the dists) so source can't shadow the installed wheel; the uncached operator build dominates wall time (~5–6 min).

Both jobs also run (build + test, no publish) on PRs touching `pyproject.toml` or this workflow; attestation is skipped on PRs.

## Verified locally
- Entry-point resolver flags a missing `main` and exits non-zero.
- Full pip install on a clean py3.13 venv + the real MCMC solve passed (both jax and numba backends).
- YAML parses; both embedded Python blocks compile.

---
> 🤖 **AI-generated disclosure:** Authored by Claude (Claude Code) at the request of @maresb. Please review before merging.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -110,13 +110,67 @@ jobs:
           python -m venv /tmp/sdist-env
           /tmp/sdist-env/bin/python -m pip install --no-deps dist/*.tar.gz
 
+  functional-test:
+    # Install the built wheel with its full dependency set and run a couple of
+    # quick checks beyond `--help`. This verifies the declared dependencies are
+    # actually installable and sufficient: importing celeri alone exercises
+    # cartopy, pymc, numba, pyproj, shapely and the rest of the stack, so a
+    # missing or under-constrained runtime dependency fails here.
+    needs: build-package
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download Distribution Artifacts
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          # The build-and-inspect-python-package action invokes upload-artifact.
+          name: Packages
+          path: dist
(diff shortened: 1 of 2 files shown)


maresb:
Updated `functional-test` to go beyond a `--help` check: it now installs the built wheel with its full dependency set and runs a real end-to-end MCMC solve (`build_model` + `solve_mcmc`) on the `wna` model, so an installable-but-insufficient dependency set fails here. Verified locally on a clean py3.13 venv with both the jax and numba backends.

Now rebased onto `main` (#474 is merged), so this targets `main` directly.

brendanjmeade:
Thanks for building this @maresb!

maresb:
No cause for concern in case this turns red. The fundamental test passes, but I'm still cleaning this up.

maresb:
Merging since this is CI-only

PR #480, opened 2026-06-28
Fix mcmc_draw crash on divergent samples (closes #475)

### LINKED ISSUE #475: minor: mcmc_draw crashes on divergent samples (averages sample_stats divergence-reason strings)
## Summary

`Estimation.mcmc_draw` crashes whenever the MCMC trace contains a **divergent sample**, with:

```
ValueError: could not convert string to float:
  'Divergence due to large energy error: 5280.9373'
```

This surfaces intermittently as a flaky CI failure in `tests/test_solve_mcmc.py::test_mcmc_selection`, but the underlying bug is in library code and will hit any real solve that diverges.

## Traceback

```
tests/test_solve_mcmc.py:197: in test_mcmc_selection
celeri/solve.py:757: in mcmc_draw
    trace_for_state = subsetted_trace.mean(dims_to_mean)
arviz/data/base.py:485: in wrapped
xarray/core/_aggregations.py:1851: in mean
  ...
xarray/computation/nanops.py:117: in _nanmean_ddof_object
numpy/_core/fromnumeric.py: in sum
E   ValueError: could not convert string to float:
    'Divergence due to large energy error: 5280.9373'
```

## Root cause

In `Estimation.mcmc_draw` (`celeri/solve.py:757`):

```python
trace_for_state = subsetted_trace.mean(dims_to_mean)
```

`subsetted_trace` is the full `InferenceData`, so `.mean()` is applied across **every** group — including `sample_stats`. When nutpie records a divergence, the divergence *reason* is stored as an **object-dtype string** in `sample_stats` (e.g. `'Divergence due to large energy error: 5280.9373'`). xarray's `nanmean` then tries to `numpy.sum` that string column and raises `ValueError`.

The mean is only consumed by `_state_vector_from_draw` (`celeri/solve_mcmc.py`), which reads **only `trace.posterior.*`** variables (`block_strain_rate`, `mogi`, `rotation`, eigen coefficients). The `sample_stats` group is never needed here.

## Why it's flaky

`test_mcmc_selection` runs a tiny unseeded solve (`tune=10, draws=10, chains=4`). Divergences are intermittent, so the failure only appears when at least one draw diverges. The same test passes on `main` at the same commit when no divergence happens.

## When it started (root-cause provenance)

The crash needs two ingredients, which landed three months apart:

1. **The latent code defect** — PR #442 (`aseyboldt/sel-chain`), merged **2026-03-20** (commit `c59d8be`). It introduced the `subsetted_trace.mean(dims_to_mean)` over the full `InferenceData` *and* the `test_mcmc_selection` test in the same commit.
2. **The trigger** — **nutpie 0.16.9** (2026-05-08) changelog: *"Handle string values in trace sample stats"*. The [sample-stats docs](https://pymc-devs.github.io/nutpie/sample-stats.html) confirm `divergence_message` is a **string** field. This is what puts the `'Divergence due to large energy error: …'` string into `sample_stats`.

This reached celeri via **PR #461** (`update-pixi-deps`, merged **2026-05-22**), which bumped **nutpie 0.16.8 → 0.16.10**, crossing the 0.16.9 boundary. (For the record: PR #468, the later `update-pixi-deps`, left nutpie/pymc/pytensor/arviz unchanged and is *not* the trigger.)

So any divergence on or after #461 (2026-05-22) triggers the crash. First observed CI reds with this signature: `cursor/upgrade-pymc-v6` (2026-05-17, a PyMC 6 branch that already had the newer nutpie), then on the pinned stack `fix-check-segment-block…` (2026-06-26) and `main` at `11b20a4` (2026-06-27). It has since recurred on #466, #472 and #479 — it will keep flaking on unrelated PRs until fixed.

## Suggested fix

Restrict the mean to the `posterior` group (the only thing `_state_vector_from_draw` reads), e.g.:

```python
trace_for_state = subsetted_trace.posterior.mean(dims_to_mean)
```

(adjusting `_state_vector_from_draw` to accept a posterior dataset, or wrapping it back into an `InferenceData` as needed). Alternatively, drop non-numeric `sample_stats` variables before aggregating.

The test should probably also seed the sampler (or assert on divergence handling) so it deterministically exercises the divergent path rather than failing at random.

---
> 🤖 Filed by Claude (Claude Code) at the request of @maresb from a CI failure traceback. Root-cause provenance added after investigating the run history and nutpie changelog.

maresb:
Fixes #475.

This is a straightforward fix for an occasional crash triggered by an interaction between nutpie and celeri due to recent changes in nutpie. This shouldn't affect any results.

---

## Problem

`Estimation.mcmc_draw` (and the estimation build in `solve_mcmc`) averaged the **whole** `InferenceData` (`subsetted_trace.mean(...)`), including the `sample_stats` group. Since **nutpie ≥ 0.16.9** stores an object-dtype divergence message there (e.g. `'Divergence due to large energy error: 5280.9373'`), selecting a divergent draw tried to take the `mean()` of that string and raised:

```
ValueError: could not convert string to float: 'Divergence due to large energy error: ...'
```

It surfaced as an intermittent failure of `test_mcmc_selection` (it only fires when a draw diverges) and flaked across #466, #472 and #479. See #475 for the full root-cause provenance.

## Fix

Average **only the `posterior` group** (the only thing `_state_vector_from_draw` reads) at both call sites. `_state_vector_from_draw` now takes the posterior `Dataset` (`posterior_point`) directly. The resulting `state_vector` is identical; we simply stop reducing groups we never use, so `sample_stats` is never touched.

No new test: this de-flakes the existing `test_mcmc_selection` (it no longer crashes when a real divergence lands), and the change is an obviously-correct refactor.

---
> 🤖 AI-assisted (Claude Code) at the request of @maresb.

### DIFF
--- a/celeri/solve.py
+++ b/celeri/solve.py
@@ -754,9 +754,9 @@ class Estimation:
             dim for dim in ("draw", "chain") if dim in subsetted_trace.posterior.dims
         ]
 
-        trace_for_state = subsetted_trace.mean(dims_to_mean)
+        posterior_point = subsetted_trace.posterior.mean(dims_to_mean)
         state_vector = _state_vector_from_draw(
-            self.model, self.operators, trace_for_state
+            self.model, self.operators, posterior_point
         )
         return replace(self, state_vector=state_vector, mcmc_trace=subsetted_trace)
 
--- a/celeri/solve_mcmc.py
+++ b/celeri/solve_mcmc.py
@@ -1780,9 +1780,8 @@ def solve_mcmc(
         compile_kwargs={"mode": backend_mode},
     )
 
-    state_vector = _state_vector_from_draw(
-        model, operators, trace.mean(["chain", "draw"])
-    )
+    posterior_point = trace.posterior.mean(["chain", "draw"])
+    state_vector = _state_vector_from_draw(model, operators, posterior_point)
     estimation = build_estimation(model, operators, state_vector)
     estimation.mcmc_trace = trace
     estimation.mcmc_start_time = mcmc_start_time.isoformat()
@@ -1795,9 +1794,14 @@ def solve_mcmc(
 def _state_vector_from_draw(
     model: Model,
     operators: Operators,
-    trace,
+    posterior_point,
 ):
-    """Build a state vector from MCMC trace using eigen coefficients.
+    """Build a state vector from a single posterior point using eigen coefficients.
+
+    ``posterior_point`` is the ``posterior`` group of an MCMC trace (an xarray
+    ``Dataset``) reduced to a single point over ``draw``/``chain`` by the
+    caller: the mean of the selected draws/chains, or one sample when a single
+    draw and chain are chosen.
 
     The state vector uses eigen coefficients (not TDE values) so that
     Estimation.predictions uses eigen_to_velocities for forward predictions.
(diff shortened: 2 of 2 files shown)


brendanjmeade (approved):
New nomenclature for me to get used to!

PR #481, opened 2026-07-01
Default coupling prior mean to 0.5

maresb:
## Summary

Changes the default MCMC coupling prior mean (`Config.mcmc_default_mesh_coupling_mean`) from `0.9` to `0.5`.

This way the default prior is minimally committal between creep and full coupling rather than biased toward full coupling.

Mesh-specific `coupling_mean` values continue to override this default, so only runs that rely on the default are affected.

## Change

- `celeri/config.py`: `mcmc_default_mesh_coupling_mean: 0.9 → 0.5`


maresb:
@brendanjmeade, this seems to me like a good idea. I might be wrong. What do you think?

brendanjmeade (commented):
@maresb This is a critical and awesome find.  Thank you!

PR #490, opened 2026-07-08
Convert CLI filenames from str to Path

jploveless:
CLI-specified filenames are written to the output `config.json` (in a run directory) as cwd-relative strings, not absolute paths. This makes building a model from the output `config.json` difficult without manual editing of the file. I'm trying to do the latter to access pre-calculated elastic kernels across a suite of models. 

This PR edits `cli.py`, converting CLI filenames from strings to absolute paths.

### DIFF
--- a/celeri/cli.py
+++ b/celeri/cli.py
@@ -323,6 +323,10 @@ def process_args(config: Config, args: argparse.Namespace):
                 if isinstance(args_val, bool) and isinstance(original_val, int | float):
                     original_val = bool(original_val)
 
+                # Convert CLI filenames from cwd-relative str to absolute Path
+                if key[-10:] == "_file_name":
+                    args_val = Path(args_val).absolute()
+
                 # Only log if the value is actually being changed
                 if original_val != args_val:
                     logger.warning(f"ORIGINAL: config.{key}: {original_val}")
--- a/tests/test_process_args.py
+++ b/tests/test_process_args.py
@@ -44,7 +44,8 @@ def test_mesh_parameters_file_name_override_reloads_mesh_params(tmp_path):
     process_args(config, args)
 
     # Both the path field AND the actual loaded parameters must reflect the alt file.
-    assert config.mesh_parameters_file_name == str(alt)
+    # Assertion of filename now removes str(alt) because CLI filenames get converted to Path (PR #490)
+    assert config.mesh_parameters_file_name == alt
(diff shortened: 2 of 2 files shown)


brendanjmeade:
@jploveless Thanks for sharing.  Can you explain the use case a bit more?  I'm wary of changing internal representations at this point, and this PR seems to be not passing a test gate that expects a string.  Could this use case be handled by a standalone script?  Happy to discuss!

jploveless:
I noticed the test failing, and it definitely scares me (as does the PR in general)! 

When an alternate filename (segment, station, etc.) is specified as a CLI option, it is a string representing a path relative to the cwd. The `config` object is updated with that relative pathname. This `config` object is then written to the output run directory, and it contains a mix of absolute paths (because in building the model, any relative paths of the `*config.json` (in that case specified relative to the path of `*config.json`) are converted to absolute) and cwd-relative paths for any alternate files specified on the CLI. This therefore makes the named files used for a particular run ambiguous, because the cwd when the model was run isn't captured anywhere in the output. 

A good alternative to changing the code could be to always specify CLI filenames as absolute paths, and I'm happy to just stick with that as practice (and I'll write a quick script just to retroactively change all of my `run/config.json` files to include all-absolute paths). 

My use case for wanting absolute paths in the `(run)/config.json` file is because I want to access the stored elastic partials for a suite of run directories. For Japan models, I want to calculate velocities due to individual meshes, in order to compare velocities due to subduction zone slip to those due to BEL slip. If there's a different way to access the stored elastic partials (rather than building a model and using that to find the hash of the cache), could you let me know? I think it's probably something you've done for the WNA postseismic transient stuff you've done.

brendanjmeade:
@jploveless I totally understand this now!  This seems completely sensible and like a good idea.  It legitimately seems like "fixing" the test is the right thing to do because it was only ever a string because I was too lazy to ever cast it to Path.  @jploveless Do you want to modify the test?  Alternatively, I'm happy to do a Claude-driven fix, which will be more careful than I!

jploveless:
Ha! This approach of editing the test feels totally backwards to me, but in this case, I think it's as simple as removing the cast to `str` that was done in assertion of the alternate mesh filename. `test_process_args.py` is now passing locally for me; hopefully it passes remotely, too!

brendanjmeade:
@jploveless It may feel a bit strange, but you've now _improved_ the test!

PR #494, opened 2026-07-11
Add option to abort stuck chains

maresb:
An implementation of @aseyboldt idea to use nutpie's non-blocking sampler to detect and abort stuck chains when the user sets `mcmc_drop_stalled_chains = True`.

I added an arguably over-engineered heuristic allowing the watchdog to immediately abort once all non-stuck chains have finished, so that you don't have to wait for any arbitrary thresholds. (A chain is considered stuck if another chain has completed, and the pending chain's time-weighted median step duration is 5x slower than the completed chain's 90th percentile.)

### DIFF
--- a/celeri/config.py
+++ b/celeri/config.py
@@ -232,6 +232,24 @@ class Config(RelativePathSerializerMixin, BaseModel):
     mcmc_backend: Literal["numba", "jax"] = "jax"
     """Backend to use for MCMC computations."""
 
+    mcmc_drop_stalled_chains: bool = False
+    """Abort and drop stalled MCMC chains once the other chains finish.
+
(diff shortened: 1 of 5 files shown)


brendanjmeade:
@maresb Wow!  I'll try running this with a large number of chains (to make a stuck chain likely) and the WNA model this week.  I'll set `mcmc_drop_stalled_chains` to  `True` in the command file.  If this works and the batched matrix-vector multiplies reduce memory usage (https://github.com/brendanjmeade/celeri/pull/487, https://github.com/brendanjmeade/celeri/pull/488), the path million+ data models might be open.

brendanjmeade:
Merging now to test tomorrow.

Issue #495, opened 2026-07-16
arviz syntax?


brendanjmeade:
Trying to run with https://github.com/brendanjmeade/celeri/pull/494 and hitting

```
 celeri-solve ./data/config/wna_08_100k_nostrain_bel_config.json --solve_type mcmc --mcmc-backend jax --mcmc-chains 4
/Users/meade/Desktop/celeri-org/celeri/.pixi/envs/default/lib/python3.13/site-packages/arviz/__init__.py:110: MigrationWarning: arviz.InferenceData is no longer available on the arviz package; ArviZ now uses xarray's DataTree for the same role. See the migration guide: https://python.arviz.org/en/latest/user_guide/migration_guide.html#datatree
  warnings.warn(
Traceback (most recent call last):
  File "/Users/meade/Desktop/celeri-org/celeri/.pixi/envs/default/bin/celeri-solve", line 4, in <module>
    from celeri.scripts.celeri_solve import main
  File "/Users/meade/Desktop/celeri-org/celeri/celeri/__init__.py", line 89, in <module>
    from celeri.solve_mcmc import solve_mcmc
  File "/Users/meade/Desktop/celeri-org/celeri/celeri/solve_mcmc.py", line 24, in <module>
    import pymc as pm
  File "/Users/meade/Desktop/celeri-org/celeri/.pixi/envs/default/lib/python3.13/site-packages/pymc/__init__.py", line 50, in <module>
    from pymc import _version, gp, math, ode, sampling
  File "/Users/meade/Desktop/celeri-org/celeri/.pixi/envs/default/lib/python3.13/site-packages/pymc/sampling/__init__.py", line 17, in <module>
    from pymc.sampling.deterministic import compute_deterministics
  File "/Users/meade/Desktop/celeri-org/celeri/.pixi/envs/default/lib/python3.13/site-packages/pymc/sampling/deterministic.py", line 20, in <module>
    from pymc.backends.arviz import apply_function_over_dataset, coords_and_dims_for_inferencedata
  File "/Users/meade/Desktop/celeri-org/celeri/.pixi/envs/default/lib/python3.13/site-packages/pymc/backends/__init__.py", line 72, in <module>
    from pymc.backends.arviz import predictions_to_inference_data, to_inference_data
  File "/Users/meade/Desktop/celeri-org/celeri/.pixi/envs/default/lib/python3.13/site-packages/pymc/backends/arviz.py", line 31, in <module>
[...]

PR #496, opened 2026-07-17
Update Pixi platform schema

maresb:
## Summary

- move the macOS 13.3 virtual-package requirement from the deprecated
  `[system-requirements]` table to the `osx-64` and `osx-arm64` platform entries
- preserve the existing Linux, macOS, and Windows platform coverage
- leave `pixi.lock` unchanged because the migration is semantically equivalent

## Why

Pixi 0.72 deprecates workspace-level system requirements in favor of declaring
virtual packages directly on rich platform entries. The old schema still parses
for compatibility but emits a warning.

## Impact

The manifest no longer emits the deprecation warning with current Pixi. Both
macOS targets continue to resolve with macOS 13.3 as their minimum virtual
package, with no dependency-resolution changes.

## Validation

- `pixi 0.72.2 workspace platform list --manifest-path pixi.toml`
- `pixi 0.72.2 lock --manifest-path pixi.toml --check`
- repository commit hooks

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -4,7 +4,12 @@
 
 [workspace]
 name = "celeri"
-platforms = ["linux-64", "osx-64", "osx-arm64", "win-64"]
+platforms = [
+    "linux-64",
+    { platform = "osx-64", macos = "13.3" },
+    { platform = "osx-arm64", macos = "13.3" },
+    "win-64",
+]
 channels = ["conda-forge"]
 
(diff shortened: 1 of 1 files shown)


maresb:
@brendanjmeade this should be ready to merge. It doesn't do anything against #495 but it removes the deprecation warnings that may tempt people or agents to attempt weird things.

PR #504, opened 2026-08-18
Enforce CCW circulation in segmesh

jploveless:
The sign conventions used in `celeri` assume counterclockwise node ordering for meshes (especially the finicky projection of relative block motion onto triangle geometries to yield slip rates). When meshing faults using `segmesh.py`, this ordering may be reversed, depending on how the ordered block coordinates are circulated. This has an impact on the sign of reported dip-slip kinematic rates, and therefore it may force slip deficit rates to adopt that same sign if dip-slip coupling constraints are applied. 

This only impacts dipping meshes, and it may not impact every dipping fault; it depends on the ordering of the segment coordinates used to create the mesh. But in previously generated segmeshes, elements that are ordered clockwise, the output sign of kinematic dip-slip is backwards, and for affected faults with coupling constraints, the output sign of the dip-slip deficit is also backwards. 

My initial test suggests that this has negligible impact on the model fit, because the sign of kinematic and elastic dip-slip are impacted equally, and therefore coupling constraints should be correctly imposed/satisfied. `cutde` is, I believe, working just fine, as it's essentially using the "bottom" face of clockwise elements rather than the conventional "top" for counterclockwise elements.   

Running this version of `segmesh.py` will print a report which faults get their circulation corrected (to be CCW). The correction is applied to the segment perimeter coordinates, which propagates into the circulation of the derived elements.

### DIFF
--- a/celeri/scripts/segmesh.py
+++ b/celeri/scripts/segmesh.py
@@ -342,7 +342,17 @@ def main():
             filename = mesh_dir / f"{seg_file_stem}_segmesh{mesh_idx}.msh"
 
             # Combined coordinates making a continuous perimeter loop
-            all_coords = np.vstack((top_coords, np.flipud(bottom_coords)))
+            # Check here for circulation direction. We want a positive z-component for the cross product of corners
+            vec1 = top_coords[-1, :] - top_coords[0, :]
+            vec2 = bottom_coords[-1, :] - top_coords[-1, :]
+            normal_vector = np.cross(vec1, vec2)
+            if (
+                normal_vector[2] < -1e-10
+            ):  # Negative cross-product; we're not worried about vertical faults
+                print("neg circulation")
+                all_coords = np.vstack((np.flipud(top_coords), bottom_coords))
+            else:
+                all_coords = np.vstack((top_coords, np.flipud(bottom_coords)))
             # Number of geometric objects
             n_coords = np.shape(all_coords)[0]
             n_surf = int((n_coords - 2) / 2)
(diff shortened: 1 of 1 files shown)


brendanjmeade:
This is fantastic @jploveless.  Do you think this would have an impact on dipping structures where we specify the bounds of the elastic component of slip rather than the coupling bounds?

jploveless:
Shoot, that's a really good question, and unfortunately I do think that's going to have an impact on the solution, because the constraint is specified with the sign that implies the desired slip-sense. For example, using a positive constraint to force a mesh to have reverse slip would, I believe, impose normal slip on a CW-circulation mesh, but this wouldn't be evident in the output sign of slip (i.e. it'd be reported as positive but that means normal). 

If the bounds are symmetric about 0, the solution should be fine but again the reported slip rates would be backwards in sign.

I don't have a copy of the files corresponding to a pre-segmeshed WNA run; the current repo just has the already-segmeshed versions. But running this new version of `segmesh.py` will report the indices of any meshes that get corrected to CCW.

brendanjmeade:
Thanks, @jploveless. I'll get ready to rerun some of WNA with this.  It also prompts me to have Claude run a full adversarial audit of celeri.

jploveless:
Rerunning WNA may not be critical, if none of the dipping meshes have CW circulation. Running the new segmesh first to see what gets reported as a CW mesh (that gets flipped internally) will indicate how many faults may be problematic. Faults that were meshed CW _and_ have elastic DS constraints applied are, I believe, the only problematic cases.

PR #507, opened 2026-08-26
Make MCMC mesh and segment outputs report posterior means consistently

### LINKED ISSUE #506: Output mean posterior elastic TDE slip rates in `model_meshes.csv`
For an MCMC run, `model_meshes.csv` includes mean posterior values for coupling and kinematic slip rates, but not elastic slip rates. This is a minor discrepancy but inconsistent with what is written to the individual mesh HDF files. 

These outputs are gathered in the `mesh_estimate` DataFrame in [`solve.py`](https://github.com/brendanjmeade/celeri/blob/1779fc2742b07a348765efab18220b245b6b58b8/celeri/solve.py#L167), then written in [`output.py`](https://github.com/brendanjmeade/celeri/blob/1779fc2742b07a348765efab18220b245b6b58b8/celeri/output.py#L228).

brendanjmeade:
Fixes #506.

## Problem

For MCMC runs the three output files disagreed about the same mesh:

| | `model_meshes.csv` (before) | HDF5 (before) |
|---|---|---|
| `strike_slip_rate` / `dip_slip_rate` | posterior-mean elastic field **projected onto the truncated eigenmode basis** (`V Vᵀ mean(elastic)`), not `mean(elastic)` | same projection |
| `*_kinematic` | Gaussian-smoothed kinematic rate (0.25° default), which the sampler never used | unsmoothed |
| `*_coupling` | posterior mean of `coupling_*` | ratio `projected elastic / unsmoothed kinematic` (blows up where kinematic ≈ 0) |

`model_segment.csv` uncertainty columns were NaN for MCMC runs even though the posterior std of `segment_slip_rate` is in the trace.

## Changes

- **`Estimation.mesh_slip_fields(mesh_idx)`** is the single source for the per-mesh rate / kinematic / coupling arrays. For MCMC: posterior means of `elastic_*` and `coupling_*`, and the unsmoothed kinematic rate the sampler used (equal to its posterior mean, since it is linear in the rotation). Non-MCMC solvers keep their previous semantics. `model_meshes.csv` and the HDF5 mesh datasets are both built from it, so they now agree. (Side effect: for non-MCMC eigen runs the HDF5 kinematic/coupling are now the smoothed versions the CSV already had.)
- **Segment uncertainties** for MCMC come from the posterior std of `segment_slip_rate`.
- **`_state_vector_from_draw`** projects the sampled `elastic_*` field for both parameterizations. Previously elastic mode used the raw `elastic_eigen_*` coefficients, which drop the prior-mean offset and bound transform; a no-op for coupling mode and unbounded elastic mode with zero mean.
- Posterior mean is computed once per estimation and tolerates single-draw subsets from `mcmc_draw()`.

## Verification

- Against an existing Cascadia coupling-mode run: CSV rate/coupling == posterior mean (1e-16), kinematic == posterior mean (1.5e-6, float32 sampler), all six HDF5 datasets == CSV, segment uncertainties == posterior std (CSV rounding), station predictions == posterior `mu`.
- New shared assertion helper in `tests/test_output_files.py` checks CSV == HDF5 == posterior (means, stds, `mu`) in the existing coupling-mode MCMC test and a new bounded-elastic-mode test. `tests/test_output_files.py`, `tests/test_solve_mcmc.py`, `tests/test_serialize.py` pass; ruff clean.

### DIFF
--- a/celeri/output.py
+++ b/celeri/output.py
@@ -113,56 +113,31 @@ def write_output(
                 data=1,  # type: ignore[arg-type]
             )
 
-            if estimation.tde_rates is not None:
-                tde_ss_rates = estimation.tde_strike_slip_rates
-                tde_ds_rates = estimation.tde_dip_slip_rates
-                if tde_ss_rates is not None and tde_ds_rates is not None:
-                    grp.create_dataset(
-                        f"/meshes/mesh_{i:05}/strike_slip/{0:012}",
-                        data=tde_ss_rates[i],
-                    )
-                    grp.create_dataset(
-                        f"/meshes/mesh_{i:05}/dip_slip/{0:012}",
-                        data=tde_ds_rates[i],
-                    )
-                    grp.create_dataset(
-                        f"/meshes/mesh_{i:05}/tensile_slip/{0:012}",
-                        data=np.zeros_like(tde_ds_rates[i]),
-                    )
-
-                tde_ss_kinematic = estimation.tde_strike_slip_rates_kinematic
-                tde_ds_kinematic = estimation.tde_dip_slip_rates_kinematic
-                if tde_ss_kinematic is not None and tde_ds_kinematic is not None:
-                    if i in tde_ss_kinematic and i in tde_ds_kinematic:
-                        grp.create_dataset(
-                            f"/meshes/mesh_{i:05}/strike_slip_kinematic/{0:012}",
-                            data=tde_ss_kinematic[i],
-                        )
-                        grp.create_dataset(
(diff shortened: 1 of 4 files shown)


jploveless:
Really nice! I think this also solves a previous problem I had noticed where meshes with elastic dip-slip rates were constrained were getting written as all zero. I can't quickly find the model in which that happened, but a newly tested elastic slip rate constraint model correctly outputs the mean posterior rates. 

The table helps to highlight the discrepancies, and it's great that all outputs are now aligned.

brendanjmeade:
@jploveless Claude is simply great at this type of work!

PR #513, opened 2026-09-01
Seed the MCMC output tests and size the segment tolerance to the float32 bound

brendanjmeade:
#510's post-rebase CI failed the segment consistency assertion again, through the tolerance added in #512 (1/837 elements, abs diff 7.7e-4 on a ~1.2 mm/yr rate). That failure shape shows the #512 sizing used the wrong error model: the float32 round-off in the sampler's `segment_slip_rate` does not scale with the resulting rate, so `rtol` cannot cover it. It scales with the operand terms — the rotation cross products are up to ~4000 mm/yr per row (median ~100) and cancel down to rates of ≤100 mm/yr, so the round-off ceiling is the same for a 1 mm/yr rate as for a 70 mm/yr one.

Computing the bound directly for the WNA operator at solution-scale rotations:

```
(nnz + 2) * eps_f32 * (|A| @ |x|)   ->  max 3.6e-3 mm/yr, median 9.3e-5
```

Every miss observed so far (5.1e-4, 5.2e-4, 7.7e-4, 1.12e-3) sits inside this ceiling; the median explains why only 1–7 of 837 elements ever trip.

Two changes:

1. **Size the tolerance to the computed bound**: `atol=5e-3` (≈1.4× the ceiling), dropping the structurally wrong `rtol`. A genuine output inconsistency — wrong basis, smoothed vs unsmoothed kinematic, dropped bound transform — shows up at 0.05+ mm/yr, 10× above this, so the assertion keeps its power.

2. **Seed the two MCMC solves** (`seed: 42`). With 2 tune / 2 draws the sampler is nowhere near converged, and an unseeded wild draw can inflate the float32 operand scale past any fixed tolerance — no static `atol` is safe against that tail. Seeding makes each CI run deterministic per environment, ending the flake class rather than shrinking its probability.

Verified locally: both MCMC output tests pass and produce identical results across repeated seeded runs.

After this merges, #510 needs one more `@dependabot rebase`. #511 is already green from its first rebase and can merge as-is.

https://claude.ai/code/session_01MyHbQgwRx7p3ietwHwnp8A

### DIFF
--- a/tests/test_output_files.py
+++ b/tests/test_output_files.py
@@ -45,12 +45,12 @@ def _assert_mcmc_outputs_consistent(estimation, run_dir):
                 assert_allclose(kinematic_rate, kinematic[component::2], rtol=1e-6)
                 kinematic_var = f"kinematic_{i}_{kind}"
                 if kinematic_var in posterior_mean:
-                    # The sampler evaluates its operators in float32
+                    # float32 boundary: same operand-scale round-off as the
+                    # segment rates below
                     assert_allclose(
                         kinematic_rate,
                         posterior_mean[kinematic_var].values,
-                        rtol=1e-4,
-                        atol=1e-4,
+                        atol=5e-3,
                     )
                 coupling_var = f"coupling_{i}_{kind}"
                 if coupling_var in posterior_mean:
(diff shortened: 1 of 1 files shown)


brendanjmeade:
The first CI round proved the seed works as intended: the legacy job failed *deterministically* on the next marginal assertion — the station `mu` check (`atol=1e-3`) missed by 1.18e-3, matching the operand-scale ceiling for the rotation cross products (~`5·eps_f32·4000 ≈ 1.2e-3`). The second commit audits every assertion in `_assert_mcmc_outputs_consistent` by error class: checks comparing identical computations (mesh rate, coupling, HDF5, kinematic-vs-CSV, segment std) stay tight; the three float32-boundary checks (segment rates, station mu, kinematic-vs-posterior) all get `atol=5e-3`. Verified locally on both stacks: the two seeded default-env tests pass, and the full legacy-job test list passes (33 tests, including the one that failed here).

PR #517, opened 2026-09-16
Revert the vertex reordering that flipped dip-slip signs, and make the kinematic smoothing configurable

brendanjmeade:
## Summary

PR #515 did two things in one commit (`c56f493`): it fixed the triangle strike/dip computation (a local east-north-up frame instead of raw lon/lat legs) and it **reordered the vertices of every triangle whose normal points down**. The second part was described as normalising "one reversed Cascadia triangle". It was much more than that: `segmesh` writes every ribbon mesh with downward normals, so the reordering reversed every triangle of every mesh the group makes and **changed the stored sign of every dip-slip quantity on those meshes** (kinematic rates, elastic rates, the state vector, the MCMC `elastic_*_ds` fields, and the meaning of asymmetric `elastic_constraints_ds` bounds). The physics was unchanged, because the kinematic factor `1/cos(dip)` and cutde's dip-slip column flip together, but changing a sign convention was never a goal and the effect was not measured or reported.

This PR **reverts the reordering** and keeps the strike/dip fix. It also explains the "stripes" seen on the Longitudinal Valley fault meshes after PR #515 and makes the kinematic-rate smoothing that used to hide them a configuration setting.

| Quantity | Before `c56f493` | With PR #515 | This PR |
|---|---|---|---|
| Vertex order in `mesh.verts` | As in the .msh file | Reversed on every downward-wound triangle | As in the .msh file |
| `mesh.dip` on such triangles | 90 to 180 deg | 0 to 90 deg | 90 to 180 deg |
| `mesh.strike` on such triangles | Strike + 180 deg | Strike | Strike + 180 deg (the kinematic code reduces strike mod 180) |
| Kinematic strike-slip rate | | Unchanged | |
| Kinematic dip-slip rate | Sign follows the file's winding (negative = reverse on downward-wound triangles) | Positive = reverse everywhere | Sign follows the file's winding |
| cutde strike-slip and tensile columns | | Unchanged | |
| cutde dip-slip column | | Negated on reordered triangles | As before |
| Station velocities for a given physical slip | | Identical | |
| Stored dip-slip rates, state vector, MCMC `elastic_*_ds`, asymmetric `elastic_constraints_ds` | File's convention | Opposite sign on downward-wound meshes | File's convention |
| Mixed-winding meshes (WNA mesh 48: 463 of 689 triangles down; Cascadia: 1) | Per-triangle sign inconsistent | Consistent | Inconsistent, reported by a warning naming the mesh and the counts |

Verification, on the WNA model at the state vector of its July MCMC run. On the four uniformly downward-wound meshes (15, 17, 18, 45) the kinematic dip-slip rates of the reverted code now agree with the pre-PR-515 code to 0.04 to 0.7 mm/yr, and their correlation is +0.999 where PR #515 gave -0.999; what remains is the strike/dip fix itself. Meshes with near-vertical elements (48, 74, 90) differ by more because `1/cos(dip)` amplifies the same small dip change, which is the pre-existing proposal R2-5 and not this PR. A Taiwan segmesh loads with 0 triangles reordered. The three WNA baselines that PR #515 regenerated for the reordering are regenerated again and are **byte-identical** to their versions before that regeneration, so the revert restores those numerics exactly.

## The stripes were not the strike/dip fix

The strike/dip fix does not create per-element noise: on nine meshes the new strike differs from the old by a near-constant per-mesh offset (1.2 to 6.3 deg) and dip by at most 2.4 deg, and neighbour-to-neighbour roughness of strike and dip is identical before and after. The stripes are the true per-element kinematic rates of a gmsh ribbon mesh (adjacent triangles differ in strike by up to 20 deg and in dip by up to 10 deg, and `1/cos(dip)` amplifies the dip jitter), which the MCMC sampler has always multiplied coupling by. What changed between the "before" and "after" figures is the output: until PR #507 (`fa386e8`, 2026-08-26) eigen and MCMC runs wrote the Gaussian-smoothed kinematic rate (0.25 deg kernel) and the eigen-projected elastic field; after it MCMC wrote the raw per-element rate and coupling x raw kinematic. On WNA run 84 the stored kinematic rates equal the Gaussian-smoothed rates of the old code to 0.000 mm/yr on every dipping mesh, while the raw rates differ by up to 80 mm/yr.

## Configurable kinematic smoothing

- `MeshConfig.kinematic_smoothing_length_scale` (km, Gaussian over the straight-line distance between element centroids; `None` inherits `Config.mesh_default_kinematic_smoothing_length_scale`, default 25 km; `0` disables).
- The smoothing is applied to the kinematic operator itself at build time, so the MCMC coupling model, the qp2/SQP bounds, the dense flag-2 boundary rows and every output use one smooth field (`operators.rotation_to_tri_slip_rate`); the raw per-element operator is kept as `rotation_to_tri_slip_rate_raw` and written to `model_meshes.csv` and the HDF5 mesh groups as `*_kinematic_raw`.
- Dip-slip rows are smoothed in the convention of an upward-wound element (each element's winding sign is divided out before and multiplied back after), so a mixed-winding mesh is smoothed physically; stored signs are untouched. Uniformly wound meshes (every segmesh mesh, every Japan mesh) are unaffected by this detail.
- The degree-based `iterative_coupling_smoothing_length_scale` is deprecated and converted (x 111.19 km/deg) with a warning; the new field is excluded from the elastic-operator cache key, so tuning it does not recompute the Green's functions.
- Known consequences: the smoothing spreads the `1/cos(dip)`-amplified dip-slip rate of a near-vertical element over its neighbours (audit proposal R2-5 remains the fix); operators saved before this change carry only the unsmoothed operator, and `Operators.from_disk` warns; and on a mixed-winding mesh the stored field keeps a real sign jump at the winding boundary instead of the cancelled average the old winding-blind kernel produced, which is what the new warning is for.

Measured end to end on the WNA model at the state vector of its July MCMC run, the default 25 km kernel reproduces the smooth field that run's output carried. Largest neighbour-to-neighbour jump in the kinematic dip-slip rate, in mm/yr:

| mesh | per-element (raw) | smoothed, 25 km | run 84, 0.25 deg kernel |
|---|---|---|---|
| 17 | 8.26 | 0.03 | 0.03 |
| 47 | 4.15 | 0.14 | 0.15 |
| 74 | 106.16 | 0.07 | 0.06 |
| 86 | 3.74 | 0.06 | 0.06 |
| 90 | 681.21 | 1.12 | 0.18 |
- `docs/audit-celeri-solve-2026-09.md` gets a section 8 recording the correction.

## Test plan

- [x] `tests/test_mesh_geometry.py`: file winding preserved; a reversed strip reports the same plane with dip in (90, 180] and the opposite strike; cutde strike-slip/tensile columns equal and dip-slip columns exactly negated between the two windings; mixed-winding warning
- [x] `tests/test_solve_dense.py`: raw operator baseline unchanged; smoothed operator equals the row-normalised Gaussian of the raw one; length scale 0 gives the raw operator
- [x] `tests/test_config.py`: default propagation; deprecation conversion
- [x] `tests/test_solve_dense.py`: reversing the winding of some elements leaves the smoothed dip-slip field consistent with a uniformly wound mesh; checked on the shipped Cascadia mesh (1 of 1841 triangles reversed), where the smoothed row of that element keeps its own convention (cosine +0.99 with its raw row, against -0.98 for winding-blind smoothing) while neighbour roughness still halves
- [x] `tests/test_cache.py`: changing the smoothing length scale does not change the elastic-operator cache key
- [x] `tests/test_output_files.py`: MCMC outputs report the smoothed and raw kinematic rates consistently in CSV and HDF5
- [x] CI groups `solve` (with `--arraydiff`), `optimize`, `other` green locally with `MPLBACKEND=Agg`


jploveless:
@brendanjmeade Just checking in about the triangle winding order issues. There is now a report about which triangles have downward normals, but with this reverting I'm lost as to what actually changed. Perhaps the best thing is to do what I think we did in Blocks, and that is check for downward normals upon loading a mesh and just swap the second and third columns of the `verts` array to enforce counterclockwise winding and therefore upward normals. I'm curious about Claude's statement that every mesh coming out of `segmesh` is clockwise-wound; that is possible depending on the ordering of the perimeter segment coordinates, but it's something I think I fixed in a recent commit. 

I just wrote a little helper that writes a new `*_ccw.msh` file with adjusted node ordering for downward-normal elements, but I haven't PRed because I wanted to get your thoughts about whether node reordering should be done automatically (if need be) upon reading each mesh in a `celeri-solve` run, or it should be intentionally left as a distinct function that produces a corrected `.msh` file.

brendanjmeade:
@jploveless Honestly, I'm lost too and I want to get this figured out!  I think it's sensible to have a dedicated interactive Claude session to work on and understand this clearly.  I'm out of tokens until Friday (what a strange world!).  Is the little helper (love this phrase) a preprocessing script?

jploveless:
@brendanjmeade I just uploaded my little helper as #522, with some comments about potential modifications (and some further questions about sign conventions, not entirely related to element winding but spawned as a result of thinking about it).

PR #519, opened 2026-09-22
Fall back to the mesh-parameters directory when no meshes exist

minlee-tw:
celeri-segmesh raised IndexError on any model whose mesh-parameters file lists no meshes, because the output directory was taken from model.meshes[0].file_name.parent.


brendanjmeade:
Thanks for finding and fixing this @minlee-tw!  It's a nice, clean PR and I'm merging it.

PR #521, opened 2026-09-22
Write oversized row indices as datasets instead of attributes

minlee-tw:
write_output failed with "object header message is too large" on models with many stations, because row indices were stored as HDF5 attributes. HDF5 limits a single object-header message to 64 KB, which is about 8000 int64 values, and a run with 48,781 stations needs roughly 380 KB for station_index alone.


brendanjmeade:
@minlee-tw Thanks for identifying this issue and creating this PR.

Two notes:
- I haven't hit this issue even on the 100k runs that I've done.  Are you seeing crashes on your 50k runs?
- I've had Copilot check compatibility with https://github.com/brendanjmeade/parsli, and this change won't break anything there.

